PR #49, opened 2015-04-23
fixed an error with callbacks returned from wrapped Lua coroutines

kmike:
I've got an obscure LuaError when trying to use a callback yielded from a coroutine when a debug hook is enabled. 

It boils down to `resume_lua_thread` implementation which doesn't match coroutine.resume. Check the tests: `test_coroutine_yields_callback_debug_hook_nowrap` which uses coroutine.resume passes in lupa 1.1, but `test_coroutine_yields_callback_debug_hook` fails.

This is fixed by bringing implementation closer to [auxresume](http://www.lua.org/source/5.2/lcorolib.c.html). 

The main difference is that now a Lua function yielded by a coroutine gets 'main' state (not coroutine's state) as its state. This makes test pass in Lua 5.2.

### DIFF
--- a/lupa/tests/test.py
+++ b/lupa/tests/test.py
@@ -1425,19 +1425,23 @@ class TestLuaCoroutinesWithDebugHooks(SetupLuaRuntimeMixin, unittest.TestCase):
                 coroutine.yield()
             end
         ''')
-        def _check():
-            dct = {}
+        def _check(dct):
             coro = self.lua.eval('func').coroutine(dct)
             next(coro)
             cb = dct['cb']
             self.assertEqual(cb(), 123)
 
         # sending a callback should work without a debug hook
-        _check()
+        _check({})
 
         # enable debug hook and try again
         self._enable_hook()
-        _check()
+
+        # it works with a Lua table wrapper
+        _check(self.lua.table_from({}))
+
+        # FIXME: but it fails with a regular dict
+        # _check({})
 
     def test_coroutine_sets_callback_debug_hook_nowrap(self):
         resume = self.lua.eval("coroutine.resume")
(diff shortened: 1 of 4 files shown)


lupa/tests/test.py
```diff
@@ -1372,6 +1372,101 @@ def test_coroutine_while_status(self):
...
+        _check({})
+
+        # enable debug hook and try again
+        self._enable_hook()
+
+        # it works with a Lua table wrapper
+        _check(self.lua.table_from({}))
+
+        # FIXME: but it fails with a regular dict
+        # _check({})
```
kmike: This test fails because dct['cb'] is assigned a _LuaFunction value when dct is a Python object accessed from Lua. This _LuaFunction has _state set to coroutine's state, and it doesn't work when a later code tries to call it.

It works if we pass `lua.table_from({})` instead of a regular Python dict because in this case both table wrapper and _LuaFunction object are created in Python code, and so they get main state as their _state.

Something similar happens when `lua.eval("coroutine.resume")(coro)` is used instead of `next(coro)` / `coro.send(...)` - all objects get 'main' state as their `_state`.

kmike: All checks works in LuaJIT for some reason - see https://travis-ci.org/scoder/lupa/builds/59800411. Maybe that's because debug hooks are handled differently in Lua 5.2 and LuaJIT.

kmike:
@scoder any ideas on how to improve it further? Do you think we can merge this to support at least some of the use cases?

scoder:
I'm not deep enough into Lua coroutines to come up with a good idea here. But it clearly fixes buggy behaviour and does seem to improve the current situtation, so having it in is better than waiting for a complete fix.

PR #63, opened 2015-10-27
Fixed unpacks_lua_table and getting/setting Python attributes from Lua when runtime encoding is None

kmike:
In Python 3 getting/setting Python attributes from Lua doesn't work with LuaRuntime(encoding=None) because attribute names must be unicode, not bytes. `unpacks_lua_table` doesn't work either because keyword arguments must be unicode, not bytes. 

This is fixed by using unicode for attribute lookups and keyword arguments even when runtime encoding is None. Bytes are decoded from `runtime._source_encoding`.

### DIFF
--- a/lupa/_lupa.pyx
+++ b/lupa/_lupa.pyx
@@ -47,7 +47,7 @@ except ImportError:
     import builtins
 
 DEF POBJECT = b"POBJECT" # as used by LunaticPython
-
+cdef int PY2 = PY_MAJOR_VERSION == 2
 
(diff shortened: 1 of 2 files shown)


kmike:
For the first commit tests are failing because of `encoding=None` issues; for second commit tests are failing only becuse of missing Python 3.5 (see https://github.com/scoder/lupa/pull/64).

kmike:
I'm not rebasing it to make build green because inline commit comments may be lost.

scoder:
Works for me, thanks!

PR #64, opened 2015-10-27
TST fix Travis build for Python 3.5

kmike:
By default there is no Python 3.5 installed on Travis, so tests fail.
Travis installs it if we ask Travis to run tox using Python 3.5.


kmike:
ping :)

PR #65, opened 2015-10-28
fixed calling __str__ in Python 3 when encoding is None

kmike:
In Python 3 `__str__` returns unicode, so `(<unicode>s).encode(runtime._encoding)` raises an error when encoding is None (`TypeError: expected bytes, NoneType found`). This is fixed by falling back to UTF-8: it can represent all unicode data and that's what `_LuaObject` uses in its `__str__` when encoding is None.

This PR includes all changes from https://github.com/scoder/lupa/pull/63.


scoder:
I agree that encoding is better than failing here.

PR #70, opened 2016-03-08
Add arg support to eval and execute

rakiru:
eval() and execute() can take optional args that are passed into the code chunk. This allows usage similar to loadstring("...")(args).

### DIFF
--- a/lupa/_lupa.pyx
+++ b/lupa/_lupa.pyx
@@ -242,21 +242,21 @@ cdef class LuaRuntime:
         self._raised_exception = exc_info()
         return 0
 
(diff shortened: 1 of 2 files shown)


gdude2002:
+1, would be a useful merge, and has a small set of changes

rakiru:
One thing to note is that this would remove the ability to add other args to `eval`/`execute` later on. I'd be happy to change that to require passing in a tuple of args explicitly instead of *args if desired.

scoder:
This seems to be a reasonable use case for `*args` and i don't currently see other use cases. We can still add keyword arguments later.

PR #74, opened 2016-04-06
Use tox-travis

jayvdb:
Moves the jobs onto the Travis container-based infrastructure,
and allows multiple jobs to run concurrently for much lower
elapsed build time for all platforms.

Also fixes the py32 build breakage caused by virtualenv 14.0.0

Do not activate -quiet mode by default; it can now be
customised using environment variable SETUP_OPTIONS

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,22 +1,27 @@
 language: python
-python: 3.5
+python:
+  - 3.5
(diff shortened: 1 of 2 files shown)


scoder:
Cool, thanks!

PR #76, opened 2016-04-06
Enable Travis pip caching

jayvdb:
When there is no 'install' block, Travis automatically
installs the dependencies in requirements.txt.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,11 +17,14 @@ env:
   - LUA=luajit-5.1
   - LUA=lua5.2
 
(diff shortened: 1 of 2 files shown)


jayvdb:
Here is what the warm cache builds look like:
https://travis-ci.org/jayvdb/lupa/builds/121080011
https://travis-ci.org/jayvdb/lupa/jobs/121080018#L187
(Cython is installed in ~1.4 seconds)

scoder:
nice!

PR #78, opened 2016-05-02
Don't reraise python exception in case it has been intercepted in lua

jbinary:
The discussion of the problem can be found here: http://www.freelists.org/post/lupa-dev/Python-exception-reraising-leads-to-ambiguity

I've added the warning as has been asked, but it does not look very convenient to me to actually have it because it tears the tests output apart. We could try to set up filtering though.

Please let me know if anything else is needed for this to be merged.

Thanks!

### DIFF
--- a/lupa/_lupa.pyx
+++ b/lupa/_lupa.pyx
@@ -1267,8 +1267,14 @@ cdef object execute_lua_call(LuaRuntime runtime, lua_State *L, Py_ssize_t nargs)
     # call into Lua
     with nogil:
         result_status = lua.lua_pcall(L, nargs, lua.LUA_MULTRET, 0)
-    #runtime.reraise_on_exception()
     if result_status:
+        python_errors = (
+            'error reading Python attribute/item',
+            'error during Python call',
+        )
+        for python_error in python_errors:
+            if lua.lua_tostring(L, -1).endswith(python_error):
+                runtime.reraise_on_exception()
         raise_lua_error(runtime, L, result_status)
     return unpack_lua_results(runtime, L)
 
--- a/lupa/_lupa.pyx
+++ b/lupa/_lupa.pyx
@@ -5,6 +5,8 @@ A fast Python wrapper around Lua and LuaJIT2.
 """
 
 from __future__ import absolute_import
+import warnings
+import traceback
 
 cimport cython
 
@@ -235,6 +237,19 @@ cdef class LuaRuntime:
             exception = self._raised_exception
             self._raised_exception = None
             raise exception[0], exception[1], exception[2]
+        return 0
+
+    @cython.final
+    cdef int warning_on_exception(self) except -1:
+        if self._raised_exception is not None:
+            warnings.warn(
+                u'Python exception occured\n%s' % (
+                    u''.join(traceback.format_exception(
+                        *self._raised_exception,
+                    )),
+                ),
+                RuntimeWarning,
+            )
         return 0
 
     @cython.final
@@ -1275,6 +1290,7 @@ cdef object execute_lua_call(LuaRuntime runtime, lua_State *L, Py_ssize_t nargs)
         for python_error in python_errors:
             if lua.lua_tostring(L, -1).endswith(python_error):
                 runtime.reraise_on_exception()
+            runtime.warning_on_exception()
         raise_lua_error(runtime, L, result_status)
     return unpack_lua_results(runtime, L)
 
--- a/lupa/_lupa.pyx
+++ b/lupa/_lupa.pyx
@@ -1265,7 +1265,7 @@ cdef object execute_lua_call(LuaRuntime runtime, lua_State *L, Py_ssize_t nargs)
     # call into Lua
     with nogil:
         result_status = lua.lua_pcall(L, nargs, lua.LUA_MULTRET, 0)
-    runtime.reraise_on_exception()
+    #runtime.reraise_on_exception()
     if result_status:
         raise_lua_error(runtime, L, result_status)
     return unpack_lua_results(runtime, L)
--- a/lupa/_lupa.pyx
+++ b/lupa/_lupa.pyx
@@ -240,19 +240,6 @@ cdef class LuaRuntime:
         return 0
 
     @cython.final
-    cdef int warning_on_exception(self) except -1:
-        if self._raised_exception is not None:
-            warnings.warn(
-                u'Python exception occured\n%s' % (
-                    u''.join(traceback.format_exception(
-                        *self._raised_exception,
-                    )),
-                ),
-                RuntimeWarning,
-            )
-        return 0
-
-    @cython.final
     cdef int store_raised_exception(self) except -1:
         self._raised_exception = exc_info()
         return 0
@@ -1290,7 +1277,6 @@ cdef object execute_lua_call(LuaRuntime runtime, lua_State *L, Py_ssize_t nargs)
         for python_error in python_errors:
             if lua.lua_tostring(L, -1).endswith(python_error):
                 runtime.reraise_on_exception()
-            runtime.warning_on_exception()
         raise_lua_error(runtime, L, result_status)
     return unpack_lua_results(runtime, L)
 
--- a/lupa/_lupa.pyx
+++ b/lupa/_lupa.pyx
@@ -5,8 +5,6 @@ A fast Python wrapper around Lua and LuaJIT2.
 """
 
 from __future__ import absolute_import
-import warnings
-import traceback
 
 cimport cython
 
--- a/lupa/_lupa.pyx
+++ b/lupa/_lupa.pyx
@@ -1536,8 +1536,9 @@ cdef int py_object_setindex_with_gil(lua_State* L, py_object* py_obj) with gil:
         else:
             return setattr_for_lua(runtime, L, py_obj, 2, 3)
     except:
-        try: runtime.store_raised_exception()
-        finally: return -1
+        runtime.store_raised_exception()
+        py_to_lua(runtime, L, runtime._raised_exception[1])
+        return -1
 
 cdef int py_object_setindex(lua_State* L) nogil:
     cdef py_object* py_obj = unwrap_lua_object(L, 1) # may not return on error!
@@ -1545,7 +1546,7 @@ cdef int py_object_setindex(lua_State* L) nogil:
         return lua.luaL_argerror(L, 1, "not a python object")   # never returns!
     result = py_object_setindex_with_gil(L, py_obj)
     if result < 0:
-        return lua.luaL_error(L, 'error writing Python attribute/item')  # never returns!
+        return lua.lua_error(L)  # never returns!
     return result
 
 # special methods for Lua wrapped Python objects
--- a/lupa/_lupa.pyx
+++ b/lupa/_lupa.pyx
@@ -238,8 +238,13 @@ cdef class LuaRuntime:
         return 0
 
     @cython.final
-    cdef int store_raised_exception(self) except -1:
-        self._raised_exception = exc_info()
+    cdef int store_raised_exception(self, lua_State* L, bytes lua_error_msg) except -1:
+        try:
+            self._raised_exception = exc_info()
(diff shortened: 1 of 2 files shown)


lupa/_lupa.pyx
```diff
@@ -1267,8 +1267,14 @@ cdef object execute_lua_call(LuaRuntime runtime, lua_State *L, Py_ssize_t nargs)
     # call into Lua
     with nogil:
         result_status = lua.lua_pcall(L, nargs, lua.LUA_MULTRET, 0)
-    #runtime.reraise_on_exception()
     if result_status:
+        python_errors = (
+            'error reading Python attribute/item',
+            'error during Python call',
+        )
```
scoder: What is the purpose of this whitelist?

lupa/tests/test.py
```diff
@@ -754,6 +754,28 @@ def test():
...
+    def test_reraise_pcall(self):
+        function = self.lua.eval(
+            'function(p) local r, err = pcall(p); return r, err end'
+        )
+        self.assertEqual(
+            function(lambda: 5/0),
+            (False, "error during Python call")
+        )
+
+    def test_lua_error_after_intercepted_python_exception(self):
```
scoder: This test fails in Travis.

scoder:
This change triggers lots of warnings in the test suite now and breaks a couple of tests in different Python versions. I'm also failing to understand why it does what it does. Could you enlighten me?

scoder:
For reference, here is my original comment on the mailing list:

I think the 'expected' behaviour would be to only call "reraise_on_exception()" if Lua returns an error in "result_status". However, that degrades the case where users just get it wrong and drop a pending Python exception. We could log a warning when a Python exception is found and clear the exception, but that's prone to shadowing bugs in user code.

It's worth a try, though. Could you write up a patch with the necessary tests and send a pull request? "Necessary tests" meaning that the new behaviour is tested, but that the tests also assure that exceptions correctly propagate through unprotected calls inside of Lua.

lupa/_lupa.pyx
jbinary: that's for the "test_lua_error_after_intercepted_python_exception", we only have to reraise in case it was a python exception, all exceptions in lua must raise LuaError instead.

jbinary:
Of course it adds a lot of warnings because it now makes a warning for each intercepted python exception, I said about this in my initial post in this pull request, I'd add some flag to the LuaRuntime to show the warnings and would supress them by default.

For the rest it's almost what you suggested to do and the "test_lua_error_after_intercepted_python_exception" explains the rest: if we intercepted python exception but then caught a failure in lua, there is still no sense to reraise.

lupa/tests/test.py
jbinary: says ok here: https://travis-ci.org/scoder/lupa/jobs/127188463
in which exactly configuration?

jbinary:
Is it not clear still?

lupa/_lupa.pyx
scoder: Looking at error message texts seems like a hack to me. Isn't there a better way to do what you want to achieve here?

jbinary: Yeah, I thought about it, but the issue is that I have to know if the error is Python related or Lua-related, and thus I have to distinguish those errors somehow. The only way I can suggest instead is to use some object instead of string in luaL_error with some code attribute that would signal that this is a python-related error (can we actually use python exception there as it's a usual python object actually? then we would not even need to store the last catched python exception).

But it was tricky for me to decide to use it because it's a major change and may affect error handling by tests/other projects depending on lupa.

jbinary:
Any updates?

scoder:
Ok, so I *really* don't like all those warnings. We are raising exceptions here, so we should see the exceptions, not warnings. And if we see an exception, then why additionally issue a warning?

If someone uses `pcall()` in Lua code, then the intention is clearly to catch all errors and specifically to not let any exceptions propagate. If someone wants to see the traceback despite using pcall(), then I'd rather add a simple way to call `traceback.print_exc()`. And maybe also something that properly reraises the Python exception after catching it with a `pcall()`. Or even something that correctly propagates both Lua and Python errors, depending on which has occurred, so that Lua code can just Do The Right Thing easily by calling some `python.*()`function, e.g. `python.reraise(alternative_lua_error_message)`.

I also don't like the string comparisons. But I understand the problem of passing the information through Lua which error actually lead to the final termination of the call, a Python error or a Lua error (which would take precedence over any Python errors it might have shadowed).

As far as I understand it, the error messages are normal Lua strings. Can't we pre-instantiate these message strings as constants in the Lua runtime, and then compare the message objects by identity instead of equality? Or would you see any other way to make this safe and explicit?

jbinary:
> Ok, so I really don't like all those warnings. We are raising exceptions here, so we should see the exceptions, not warnings. And if we see an exception, then why additionally issue a warning?

I personally don't like them too, I just implemented it to comply to these recommendations: http://www.freelists.org/post/lupa-dev/Python-exception-reraising-leads-to-ambiguity,5

> If someone uses pcall() in Lua code, then the intention is clearly to catch all errors and specifically to not let any exceptions propagate. If someone wants to see the traceback despite using pcall(), then I'd rather add a simple way to call traceback.print_exc(). And maybe also something that properly reraises the Python exception after catching it with a pcall(). Or even something that correctly propagates both Lua and Python errors, depending on which has occurred, so that Lua code can just Do The Right Thing easily by calling some python.*()function, e.g. python.reraise(alternative_lua_error_message).

python.eval('raise') should be enough, I suppose? I would not overcomplicate this thing, it's obvious that if you really really need the traceback, you'll be able to wrap python code accordingly, but usually you just need to make sure you'll get control after something went wrong in Python code. Currently, in my implementation of prosodypy, if some error occurs in 3rd party Python module, it's being logged in the prosody log including the traceback with no problems: https://github.com/jbinary/prosodypy/commit/7c3f1177b99ef43e920e2ee774f83b56367f8bce

> I also don't like the string comparisons. But I understand the problem of passing the information through Lua which error actually lead to the final termination of the call, a Python error or a Lua error (which would take precedence over any Python errors it might have shadowed).

Me either, and I don't know what is the best practice in Lua to distinct the errors. You actually can use any object as an error, it may not be a string, so we could raise some object (even derived from Python exception, I think) and then identify what does that object mean exactly, but I was afraid that it could break compatibility to some code that has been already written to work with lupa and decided to go with the safest way which also does not seem to be  dangerous, only not very expressional. Probably, there is a way to make those object to behave like strings also, so it would not break anything, but I'm really not sure if such a complication worth it. Since embedded lua modules raise strings, I think it should be fine for lua in toto?

> As far as I understand it, the error messages are normal Lua strings. Can't we pre-instantiate these message strings as constants in the Lua runtime, and then compare the message objects by identity instead of equality? Or would you see any other way to make this safe and explicit?

Yes, I could do so, if everything else is fine now.

scoder:
> python.eval('raise') should be enough, I suppose?

Might be, except that "raise" isn't an expression. But since we have to
implement the "detect Python exception and reraise, otherwise raise Lua
error" case anyway, why not expose a Lua-ish version of that as helper for
users?


> I don't know what is the best practice in Lua to distinct the errors. You actually can use any object as an error, it may not be a string, so we could raise some object (even derived from Python exception, I think) and then identify what does that object mean exactly, but I was afraid that it could break compatibility to some code that has been already written to work with lupa and decided to go with the safest way which also does not seem to be  dangerous, only not very expressional. Probably, there is a way to make those object to behave like strings also, so it would not break anything, but I'm really not sure if such a complication worth it. Since embedded lua modules raise strings, I think it should be fine for lua in toto?

My guess is that there's at least some code out there (for Lua or Lupa)
that depends on errors being printable strings, if only because Lua does it
that way by default. But Lupa's wrapped Python objects implement at least
stringification, that might be enough.

I find the idea of simply passing the wrapped Python exception through as
Lua error objects quite tempting. In Python 3, exceptions properly
reference their own tracebacks, i.e. exception instances are fully
self-contained and can simply be re-raised on Lua exit into Lupa.

They aren't self-contained in Py2, though. We could fix that (only in Py2)
by checking if sys.exc_info()[1] is the exception we received (by
identity), and then executing a plain "raise", and otherwise raising
explicitly the exception we received.

Would be great if you could try this.

jbinary:
So I gave it a try and updated the pull request. I removed the try block when storing original exception to the runtime as it seemed redundant to me, tell me if anything is wrong with that.

Another issue bothering me is that there are many places where we still error with a string, we should probably move py_to_lua call to store_raised_exception and thus remove it from anywhere else and then replace all luaL_error calls to lua_error instead, what do you think?

lupa/_lupa.pyx
```diff
@@ -1399,12 +1402,13 @@ cdef bint call_python(LuaRuntime runtime, lua_State *L, py_object* py_obj) excep
 
 cdef int py_call_with_gil(lua_State* L, py_object *py_obj) with gil:
     cdef LuaRuntime runtime = None
+    runtime = <LuaRuntime?>py_obj.runtime
```
scoder: This can fail, that's why it was inside of the try-except.

lupa/_lupa.pyx
```diff
@@ -1399,12 +1402,13 @@ cdef bint call_python(LuaRuntime runtime, lua_State *L, py_object* py_obj) excep
...
     try:
-        runtime = <LuaRuntime?>py_obj.runtime
         return call_python(runtime, L, py_obj)
     except:
-        try: runtime.store_raised_exception()
```
scoder: This can fail but must not raise an exception, thus the try-finally with return -1.

lupa/_lupa.pyx
```diff
@@ -1507,16 +1511,17 @@ cdef int py_object_getindex_with_gil(lua_State* L, py_object* py_obj) with gil:
         else:
             return getattr_for_lua(runtime, L, py_obj, 2)
     except:
-        try: runtime.store_raised_exception()
```
scoder: Same here, this needs to be covered by try-finally.

lupa/_lupa.pyx
jbinary: Do we know why? I mean, we have to have some tests for that then because now the tests went just fine even without catching?

scoder:
Nice. That looks really good.

I also like the idea of replacing all string exception results with proper Python exceptions. Please give it a try.

lupa/_lupa.pyx
scoder: memory error. not entirely easy to test for. :)

jbinary:
Thanks, I will.

lupa/_lupa.pyx
scoder: The basic thing to remember is that raising an exception can actually fail and raise an exception, so if there is nothing around that can deal with that (e.g. because we are being called by Lua), then you'll have to be paranoid and wrap everything in bare try-excepts and try-finallys. Luckily, Cython can issue a warning if it finds a spot where an exception can be raised and it doesn't know how to propagate it.

jbinary: OK, I'll move it under try block then, no probs.

lupa/_lupa.pyx
jbinary: But this one is harder: if exception happened and we could not save it, how can we silently pass that?

scoder: Well, what is the worst thing that can happen? If the outer calling code in Lupa detects a Lua error without a Python exception returned, then it will raise a LuaError. Given that this will really only happen in extreme cases (when running out of memory), this seems good enough to me.

jbinary: We have to put something on stack for lua to raise instead of python's exception then, and I'm not sure what? "error during Python call" again?

scoder: Sure, why not. As I said, it's an extreme case. If memory is really running out, even that might fail inside of Lua.

jbinary:
So I've updated the pull request. Surprisingly, I have not had to fix any tests. So could you please take a look?

jbinary:
Any news on this one?

PR #81, opened 2016-08-24
Add lua51.dll to package for installation

gdude2002:
This is so that you don't have to have a system-wide LuaJIT DLL on Windows, which allows one to have multiple versions of Python, using multiple versions of lupa/LuaJIT, compiled for separate architectures.

Wheels work too.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,5 +1,6 @@
 
 import sys
+import shutil
 import os
 
 from glob import iglob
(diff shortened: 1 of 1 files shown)


gdude2002:
I'm not really sure why the build output suggests that Lupa is building against LuaJIT statically when it clearly isn't (at least on Windows); this is the easiest quick-fix to making this portable that I could come up with.

gdude2002:
Screenshot of this built into a wheel and running on a clean VM

scoder:
Thanks

scoder:
I made some changes to your commit. Could you verify that it still works?
I wonder why we're looking for the ".lib" file, though, if what we want is the ".dll" file. Why should it find both? Which should be preferred?

gdude2002:
I'll take a look when I can. 

No idea why we're looking for the `lib`, honestly. I just assumed that it wasn't a mistake - the dll is definitely what we want. 

It still doesn't count as static compilation by the way - but it's a lot more portable.

PR #82, opened 2016-09-15
Call python with the same state as lua was trying to call it. This wa…

jbinary:
Initial discussion is here: http://www.freelists.org/post/lupa-dev/luaeval-returns-an-extra-value-if-being-run-from-coroutine

So I've identified the problem: when lua tries to call python from another thread, it uses a different state. But if python tries to call lua back, lupa uses the initial one. The commit stores a new stack in runtime in case it has been changed. This way lua.eval works consistently across the use cases. I'm not sure if it can lead to any race conditions but I could not come up with any possible reasons for that. Also, probably, it's not too nice to update runtime._state each time, if you know a better solution, please tell.

### DIFF
--- a/lupa/_lupa.pyx
+++ b/lupa/_lupa.pyx
@@ -1399,12 +1399,22 @@ cdef bint call_python(LuaRuntime runtime, lua_State *L, py_object* py_obj) excep
 
 cdef int py_call_with_gil(lua_State* L, py_object *py_obj) with gil:
     cdef LuaRuntime runtime = None
+    cdef lua_State* stored_state
+    updated_state = False
+
     try:
         runtime = <LuaRuntime?>py_obj.runtime
+        if runtime._state is not L:
+            updated_state = True
+            stored_state = runtime._state
+            runtime._state = L
         return call_python(runtime, L, py_obj)
     except:
         try: runtime.store_raised_exception()
         finally: return -1
+    finally:
+        if updated_state:
+            runtime._state = stored_state
 
 cdef int py_object_call(lua_State* L) nogil:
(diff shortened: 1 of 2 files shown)


lupa/_lupa.pyx
```diff
@@ -1399,12 +1399,22 @@ cdef bint call_python(LuaRuntime runtime, lua_State *L, py_object* py_obj) excep
 
 cdef int py_call_with_gil(lua_State* L, py_object *py_obj) with gil:
     cdef LuaRuntime runtime = None
+    cdef lua_State* stored_state
```
scoder: I would just initialise this to NULL and store it back if it's not NULL. There's no need for an additional flag.

jbinary: Yes, sure, I tried this way first but used None instead of NULL and it obviously failed. Pushed the new one.

kmike:
Does it by chance fix a FIXME: in test_coroutine_sets_callback_debug_hook (https://github.com/scoder/lupa/pull/49/files#r29065198)?

jbinary:
> Does it by chance fix a FIXME: in test_coroutine_sets_callback_debug_hook (https://github.com/scoder/lupa/pull/49/files#r29065198)?
I'm too tired right now to look into it precisely to understand what has happened there exactly, but if I uncomment that _check({}) line, it works just fine both with this fix and without it.

kmike:
@jbinary thanks for taking a look! which Lua version are you using?

jbinary:
> libluajit-5.1.so.2

kmike:
Ok, it also passes for me in luajit, but not in Lua 5.2.

jbinary:
This way fails regardless of this patch.

scoder:
Looks good. Thanks!

PR #87, opened 2017-06-14
Include lua stacktrace when LuaError is thrown

aaiyer:


aaiyer:
Tavis CI failed because of the latest Cython not supporting Python 3.3. The changes passed Travis on all other configured Python+Lua versions.

kmike:
Could you please add a test case for this feature?
Does it work if debug module is disabled, e.g. by a sandbox?

aaiyer:
@kmike Yes, it works if debug is disable

aaiyer:
Added unit test

scoder:
Looks good to me, thanks! (I'll fix the Py2.6 errors in the test.)

PR #89, opened 2017-07-13
tie lua with lupa to fix dependency missing

### LINKED ISSUE #67: Errors while compiling Lupa
http://stackoverflow.com/questions/33645073/unable-to-build-lupa

TitanSnow:
fix #67 

originally lupa finds lua lib installed in system so that it's very possible to be missing and cause build fail

to tie lua with lupa might be a better solution

I added submodule `lua` and compile it together with lupa so that there's no need to install lua before installing lupa


scoder:
Thanks. It is a good idea (and I appreciate your efforts) to make the installation easy for simple usage by directly including a version of Lua. But I'd rather not remove the ability to build against a system installed Lua. For various reasons (e.g.  system security updates), keeping both Lua(JIT) and Lupa in separate libraries is to be preferred over static linking.

Ideally, the build would detect whether a system wide installation of Lua(JIT) can be found and use it (maybe even selecting the newest version, e.g. prefer 5.2 over 5.1 if both are available), fall back to the shipped Lua if that is not the case, and allow users to explicitly provide the Lua installation directories via setup.py command line switches. At least some of that is currently already implemented.

The next question is then whether any future binary packages of Lupa on PyPI should statically include Lua, but they probably should in order to allow a straight forward `pip install lupa` to work.

And then there's the issue that git (AFAIK) doesn't have a way to say "link the submodule to the most recent release tag", i.e. I'd have to update the repository link from time to time and publish new (binary) packages with new Lua releases. That's a bit annoying...

scoder:
Most users on Windows don't even have a C compiler installed, so shipping the Lua *sources* won't help them. And Windows doesn't have a package management system, so security updates for Lua won't happen anyway. Thus, only complete binary packages for Lupa that include Lua would solve the problem on their side. But pretty much all other systems have some form of package management, which makes Windows the special case, once again.

And I do agree that the source package should be self-contained and *ship* Lua sources (although not necessarily *use* them), and I also think that it would be good to have any binary wheels on PyPI be self-contained and readily pip installable, especially in order to support Lupa as a dependency.

TitanSnow:
That’s great. I’m forwarding to see the improvement in the future

TitanSnow:
Would you mind me holding a pypi package contains pre-built lupa wheels?

scoder:
What do you mean with "holding" ?

TitanSnow:
That means, I wanna hold a package on PyPI, with name like `lupa-binary` and so on, contains lupa pre-built wheels

scoder:
That would mean that users must depend on a different package if they want binaries. I'd rather like to avoid that and provide binary packages directly on the lupa PyPI page. I can set up Windows builds in appveyor, have done that before for other packages. Could you implement a new pull request that configures the static build in the way I described?

TitanSnow:
Sure, I'm glad to help

I will write CI script to do these works on CI: download lua sources, and build binary wheels

PR #90, opened 2017-08-02
build wheels on appveyor

TitanSnow:
this pr contains some auto build scripts for appveyor to build pre-built wheels for Windows, support:

* Python2.6 x86
* Python2.7 x86
* Python3.3 x86
* Python3.4 x86
* Python3.5 x86
* Python3.5 x64
* Python3.6 x86
* Python3.6 x64

note that x64 wheels for python3.4- are not built because of [this issue](https://github.com/appveyor/ci/issues/1699)

### DIFF
--- a/.appveyor.ps1
+++ b/.appveyor.ps1
@@ -1,0 +1,56 @@
+$ErrorActionPreference = "Stop"
+
+# consts
+$luatar = 'LuaJIT-2.0.5.zip'
+$luadir = 'LuaJIT-2.0.5'
+$pyversion = (python -c "print(__import__('sys').version)")
+if($pyversion.Contains('32 bit')){
+    $arch = 'x86'
+}else{
+    $arch = 'x64'
+}
+$vcvarsall = (Join-Path $env:VS140COMNTOOLS '..' '..' 'VC' 'vcvarsall.bat')
+
+# clean
+Remove-Item -Recurse $luadir, 'build', 'lupa.egg-info', 'backup' -ErrorAction SilentlyContinue
+if($args[0] -eq 'clean'){
+    Exit
+}elseif($args[0] -eq 'distclean'){
+    Remove-Item -Recurse 'dist', $luatar -ErrorAction SilentlyContinue
+    Exit
+}
+
+# print info
+Write-Host ('building for ' + (python --version) + ' ' + $arch) -ForegroundColor Magenta
+
+# get lua tar
+if((Test-Path $luatar) -eq $false){
+    Invoke-WebRequest 'http://luajit.org/download/LuaJIT-2.0.5.zip' -OutFile $luatar
+}
+if((Get-FileHash $luatar -Algorithm MD5).Hash -ne 'f7cf52a049d74aee4e624bdc1160b80d'.ToUpper()){
+    throw 'MD5 mismatch'
+}
+Expand-Archive $luatar -DestinationPath .
+
+# build lua
+$origin_pwd = (Get-Location)
+Set-Location (Join-Path $luadir 'src')
+& $env:ComSpec /c "call `"$vcvarsall`" $arch && msvcbuild.bat"
+Set-Location $origin_pwd
+
+# build lupa
+if(Test-Path 'dist'){
+    Move-Item 'dist' 'backup'
+}
+pip install -r requirements.txt
+pip install wheel
+python setup.py bdist_wheel
+pip install @(Get-ChildItem dist\*.whl)[0] -U
+if(Test-Path 'backup'){
+    Move-Item backup\*.whl 'dist' -ErrorAction SilentlyContinue
+    Remove-Item -Recurse 'backup'
+}
+
+# test lupa
+python setup.py test
+Exit $lastexitcode
--- a/.appveyor.ps1
+++ b/.appveyor.ps1
@@ -9,7 +9,7 @@ if($pyversion.Contains('32 bit')){
 }else{
     $arch = 'x64'
 }
-$vcvarsall = (Join-Path $env:VS140COMNTOOLS '..' '..' 'VC' 'vcvarsall.bat')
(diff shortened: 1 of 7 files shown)


lupa/_stdint.h
```diff
@@ -0,0 +1,64 @@
...
+typedef __int8 int8_t;
+typedef __int16 int16_t;
+typedef __int32 int32_t;
+typedef __int64 int64_t;
+typedef unsigned __int8 uint8_t;
+typedef unsigned __int16 uint16_t;
+typedef unsigned __int32 uint32_t;
+typedef unsigned __int64 uint64_t;
+#ifdef _WIN64
+typedef __int64 intptr_t;
```
scoder: Since we really only need `uintptr_t`, I'd rather keep this file short.

lupa/_lupa.pyx
```diff
@@ -21,7 +21,10 @@ from cpython.method cimport (
 from cpython.version cimport PY_MAJOR_VERSION
 from cpython.bytes cimport PyBytes_FromFormat
 
-from libc.stdint cimport uintptr_t
+cdef extern from '_stdint.h':
+    # old versions of msvc does not have stdint.h
+    # typedef uintptr_t to unsigned long long -- it might be incorrect, but does not matter
+    ctypedef unsigned long long uintptr_t
```
scoder: This previously used `size_t` as base type.

.appveyor.ps1
```diff
@@ -0,0 +1,80 @@
...
+run 'python setup.py bdist_wheel'
+run ('pip install ' + @(Get-ChildItem dist\*.whl)[0] + ' -U')
+if(Test-Path 'backup'){
+    Move-Item backup\*.whl 'dist' -ErrorAction SilentlyContinue
+    Remove-Item -Recurse 'backup'
+}
+
+# test lupa
+Set-Location 'lupa\tests'
+run 'python __init__.py'
```
scoder: Why switch directories here instead of running it from the base directory?

.appveyor.ps1
```diff
@@ -0,0 +1,80 @@
+$ErrorActionPreference = "Stop"
+
+function run(){
+    param($command)
+    & $env:ComSpec /c ($command + ' 2>&1')
+}
+
+# consts
+$luatar = 'LuaJIT-2.0.5.zip'
```
scoder: I'd rather not depend on a specific version in scripts but instead download the newest one.

Also, I hadn't considered yet if Lupa should ship with LuaJIT or plain Lua by default - it's probably ok since users get a generally faster system that way. Let's stick with LuaJIT here for now.

scoder:
Hmm, thanks, but I think we might have misunderstood each other. What I said was:
"""
Ideally, the build would detect whether a system wide installation of Lua(JIT) can be found and use it (maybe even selecting the newest version, e.g. prefer 5.2 over 5.1 if both are available), fall back to the shipped Lua if that is not the case, and allow users to explicitly provide the Lua installation directories via setup.py command line switches.
"""
I meant that this should be done as part of the normal setup.py execution, i.e. in Python. It's ok to have special scripts and a predefined configuration for appveyor, but things that are generally applicable to all platforms (and that can be done in Python) should be done in Python and integrate with the expected distutils build procedure.

.appveyor.ps1
TitanSnow: Run test in base directory is doing test on the `lupa` directory in base directory. I need to test whether the wheel is installed successfully so that I need to avoid directory named lupa in current directory otherwise python will import from that directory instead of globally installed lupa

scoder:
Regarding the build failure on travis, I've removed Python 3.2 from the list of tested versions on travis. Lupa probably still works there but it's no longer officially supported.

.appveyor.ps1
scoder: Ah, ok. That might be worth a comment in the script. ;)

TitanSnow: right

TitanSnow:
> I meant that this should be done as part of the normal setup.py execution, i.e. in Python. It's ok to have special scripts and a predefined configuration for appveyor

In the script `.appveyor.ps1` it does these things:

1. download lua(JIT) source -- not implementable in setup.py
2. build lua(JIT) -- hard to implement in setup.py
3. run `setup.py` to build & install
4. run test

As you could see, it just does the configuration and main build work is unchanged and still in `setup.py`

scoder:
1) seems easy enough in Python, but I agree that 2) is some effort to get right in a cross-platform way.
But what about your original attempt to include the Lua .c files directly in the distutils build? Wouldn't that work in this case?

.appveyor.ps1
TitanSnow: I specify version because of these reasons:

* How to get the newest version in script? People could look through the website to know the newest version, but it's hard to implement in script. A way is to clone the git repo of luajit, but that's unreleased and riskful
* I need to verify the downloaded file's MD5. If don't specify a version, I could not know its MD5
* lupa seems to fail tests when bundled with lua5.3 by my test

A good way is to accept arguments to let users specify which version to build

TitanSnow:
I think step 1) should not be written in setup.py because it's not and shouldn't be the job of setup.py. Users would not want setup.py to download things when they're doing installation

Statically link lua library with the C extension is a good way. I'd like to reopen #89 if you're think so. But setup.py should never do downloading -- in my opinion

scoder:
Ok, but that means that if Lupa is supposed to be usable as a dependency, then it has to be self-contained *without* downloading anything. Meaning, the Lua(JIT) sources would need to be included in the sdist. With the drawback that they would get stale over time, but could get updated with the next Lupa release.
That's mostly what #89 did, except that it made the local source build the *only* option. It should only be used if it's explicitly requested via a setup.py option, or as a fallback if a suitable local installation really cannot be found. I'd be happy to see #89 changed in that direction.

PR #91, opened 2017-08-12
tie lua with lupa to fix dependency missing (v2)

TitanSnow:
#89 v2

make the local source build an *extra* option

I use lua5.3, but it could not pass the test `test_call_str_py`. I have no idea about this. Please help


scoder:
Nice! I pushed a fix for the test.

Issue #92, opened 2017-08-18
expose lua C APIs and ``lua_State`` object


TitanSnow:
expose all or a limited subset of lua C APIs and the ``lua_State`` object to make it able to do low-level operations like syntax check

scoder:
I'm not sure I understand your use case. How is a "syntax check" a low-level operation?

But the main question is: what level of safety should that C-API provide? The LuaRuntime uses locking, for example. Should that be exposed or even enforced? Should there be access to the ``lua_State`` while it is running?

How do you envision the usage of that C-API? Would you expect external packages to compile against Lupa and use it in order to extend Lupa? And, wouldn't Lua be a better extension language here than C or Cython?

TitanSnow:
> I'm not sure I understand your use case. How is a "syntax check" a low-level operation?

`luaL_loadbuffer` returns `LUA_ERRSYNTAX` when there's syntax error. In this way I could know whether a snippet of code is wrong on syntax without running it. As you can know it's important to write a REPL because REPL should know one line user typed is already finished or not and whether to wait for more input

> The LuaRuntime uses locking, for example. Should that be exposed or even enforced?

My option is to mark these APIs as "non thread-safe" and let user lock by themselves

scoder:
> luaL_loadbuffer returns LUA_ERRSYNTAX when there's syntax error.

You currently get a `LuaSyntaxError` exception in that case. If that's not enough, I think the obvious first step would be to increase the amount of information provided by that exception, rather than stepping down to the C level. If your goal is to implement a Lua REPL on top of Lupa, that should be possible in Python, without resorting to C.

> My option is to mark these APIs as "non thread-safe" and let user lock by themselves.

Which then suggests that either the `LuaRuntime` or its lock would have to get exposed. I'd rather like to avoid that...

If the C-API is purely function based, however, it could avoid exposing the runtime all together. There could be functions `int lock_runtime(object) except -1` and `int unlock_runtime(object) except -1` that could cast and lock/unlock. And a function `lua_State* get_lua_state(object) except NULL` that would return the `lua_State*` of the `LuaRuntime` that was passed in. Simple and unsafe. And a bit ugly, but well, at least simple enough to be ugly for good. Anything else you would need?

TitanSnow:
> If that's not enough, I think the obvious first step would be to increase the amount of information provided by that exception, rather than stepping down to the C level.

[...]

PR #93, opened 2017-08-19
add ``LuaRuntime.compile()``

TitanSnow:
#92


lupa/_lupa.pyx
```diff
@@ -263,6 +263,26 @@ cdef class LuaRuntime:
             lua_code = (<unicode>lua_code).encode(self._source_encoding)
         return run_lua(self, lua_code, args)
 
+    def compile(self, lua_code):
```
scoder: I'll add a docstring.

scoder:
Thanks, looks really good!

lupa/_lupa.pyx
TitanSnow: I just forget it :smile:

PR #102, opened 2018-05-11
fix #101

### LINKED ISSUE #101: UnicodeDecodeError in pip install
It seems the setup.py cause a error when system default encoding is not utf-8.
```
    Using bundled Lua
    building without Cython
      File "<string>", line 1, in <module>
      File "C:\Users\dell\AppData\Local\Temp\pip-req-build-zth2l84p\setup.py", line 308, in <module>
        for text_file in ['README.rst', 'INSTALL.rst', 'CHANGES.rst', "LICENSE.txt"]])
      File "C:\Users\dell\AppData\Local\Temp\pip-req-build-zth2l84p\setup.py", line 308, in <listcomp>
        for text_file in ['README.rst', 'INSTALL.rst', 'CHANGES.rst', "LICENSE.txt"]])
      File "C:\Users\dell\AppData\Local\Temp\pip-req-build-zth2l84p\setup.py", line 298, in read_file
        return f.read()
    UnicodeDecodeError: 'gbk' codec can't decode byte 0x93 in position 1183: illegal multibyte sequence
```
dicuss here:https://github.com/pypa/pip/issues/4251#issuecomment-387664399

TitanSnow:
add `encoding='utf8'` to avoid other encoding on Windows

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -294,7 +294,7 @@ if cythonize is not None:
 
 
 def read_file(filename):
-    with open(os.path.join(basedir, filename)) as f:
+    with open(os.path.join(basedir, filename), encoding='utf8') as f:
         return f.read()
 
 


setup.py
```diff
@@ -294,7 +294,7 @@ def has_option(name):
 
 
 def read_file(filename):
-    with open(os.path.join(basedir, filename)) as f:
+    with open(os.path.join(basedir, filename), encoding='utf8') as f:
```
scoder: This requires a `from io import open` to support Py2.x.

PR #129, opened 2019-12-20
Can't import modules on python 3

jbinary:
use same dlopen flags on python 3 as for python 2 as absence of RTLD_NOW leads to an error: `invalid mode for dlopen(): Invalid argument` and thus it's not longer possible to import modules with lupa


scoder:
Thanks! I did almost the same thing in c78d9c63238b5535b1881f4eee54700f5a138b04, just preferring Py3 over Py2 and increasing the coverage of the `ImportError`.

Issue #121, opened 2019-09-24
Tables as kwargs for Python Classes


SoulSen:
I've seen #35 already, but that only implements it for Python functions, any ways that this could be implemented for classes?

scoder:
Could you be more specific about what you are asking?

SoulSen:
Something like this
```py
class MyClass:
   def __init__(self, **kwargs): pass 
```
```lua 
MyClass{hello="hello"}
```

scoder:
Don't the decorators work for you?

SoulSen:
`TypeError: cannot create 'cython_function_or_method' instances` 

I get the error, when one of my extensions to `discord.py` is loaded and subclasses a class, I wrapped with the decorator.

scoder:
And when you wrap the __init__ method instead of the class?

SoulSen:
I have yet to try that, I will when I find time

SoulSen:
I'm not sure if I'm doing anything wrong, but after testing it, it doesn't seem to work. 

So for my use case I'm trying to make the classes and functions of a package be compatible with named parameters. 

So I do this by doing this
```python
import discord 

def _update_classes(self, discord_pkg):
        for discord_obj in self._get_package_objs(discord_pkg):
            if inspect.isclass(discord_obj):
                for cls_object in self._get_class_objs(discord_obj):
                    try:
                        cls_object = getattr(discord_obj, cls_object)
                        setattr(discord_obj, cls_object.__name__, lupa.unpacks_lua_table(cls_object))
                    except Exception:
                        pass
```

I get this error when trying to run the bot, since I the function above is called before the bot is started.

```python
Traceback (most recent call last):
  File "C:/Users/taluk/Documents/Coding Projects/Python/Musix/musix.py", line 27, in <module>
    Musix(command_prefix='hey musix ').run('TOKEN')
  File "lupa\_lupa.pyx", line 468, in lupa._lupa.unpacks_lua_table_method.wrapper
  File "C:\Users\taluk\Documents\Coding Projects\Python\Musix\venv\lib\site-packages\discord\client.py", line 598, in run
    return future.result()
  File "C:\Users\taluk\Documents\Coding Projects\Python\Musix\venv\lib\site-packages\discord\client.py", line 579, in runner
    await self.start(*args, **kwargs)
  File "C:\Users\taluk\Documents\Coding Projects\Python\Musix\venv\lib\site-packages\discord\client.py", line 542, in start
[...]

PR #133, opened 2019-12-26
Support Python calls with `@lupa.unpacks_lua_table` decorator 

SoulSen:
This pull request now allows functions or methods decorator with `@lupa.unpacks_lua_table` or `@lupa.unpacks_lua_table_method` to be called from Python. Originally if a decorated function or method was called, it would error because of invalid parameters, though this fixes that. It addresses and fixes the problems had with #121 also.

The changes has been tested and passed all checks.

### DIFF
--- a/lupa/_lupa.pyx
+++ b/lupa/_lupa.pyx
@@ -452,7 +452,9 @@ def unpacks_lua_table(func):
     """
     @wraps(func)
     def wrapper(*args):
-        args, kwargs = _fix_args_kwargs(args)
+        kwargs = {}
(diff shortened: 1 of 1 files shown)


SoulSen:
How would overriding `LuaTable` work though? Since if I'm not mistaken but you can't add anything similar to kwargs from Lua if your already using named parameters. Could you show an example of what your thinking about?

scoder:
I meant the case of calling the function from Python code. That would allow passing keyword arguments, in addition to wherever it got a `LuaTable` object from (e.g. from a call from Lua).

PR #145, opened 2020-08-07
Python exceptions within coroutines are reraised

### LINKED ISSUE #144: lupa does not print Python error when using coroutines
If you use coroutines, lupa will not show a Python error. Here a code to reproduce:
```python
from lupa import LuaRuntime

def test():
    int("Hello")

lua = LuaRuntime()
func = lua.eval("function(test) test() end") 
co = func.coroutine(test)
co.send(None)
```
This produces the following traceback:
```
  File "./show_error.py", line 10, in <module>
    co.send(None)
  File "lupa/_lupa.pyx", line 876, in lupa._lupa._LuaThread.send
  File "lupa/_lupa.pyx", line 941, in lupa._lupa.resume_lua_thread
  File "lupa/_lupa.pyx", line 1268, in lupa._lupa.raise_lua_error
lupa._lupa.LuaError
```
As you can see, the Python error (ValueError: invalid literal for int() with base 10: 'Hello') is not shown in the traceback. This makes larger code really hard to debug.

noahcgreen:
Fixes #144.
Currently if a lua coroutine calls Python code which raises an exception, a LuaError will be raised and the original exception will be lost. This isn't consistent with the behavior of non-coroutine code, which is to reraise the exception, and losing the original exception makes it very hard to debug lua coroutines.

A new unit test `test_coroutine_reraises_python_error` has been added.

### DIFF
--- a/lupa/_lupa.pyx
+++ b/lupa/_lupa.pyx
@@ -1294,7 +1294,11 @@ cdef int raise_lua_error(LuaRuntime runtime, lua_State* L, int result) except -1
     elif result == lua.LUA_ERRMEM:
         raise MemoryError()
     else:
-        raise LuaError( build_lua_error_message(runtime, L, None, -1) )
+        error = py_from_lua(runtime, L, 1)
+        if isinstance(error, BaseException):
+            runtime.reraise_on_exception()
+        else:
+            raise LuaError( build_lua_error_message(runtime, L, None, -1) )
 
 cdef build_lua_error_message(LuaRuntime runtime, lua_State* L, unicode err_message, int n):
     cdef size_t size = 0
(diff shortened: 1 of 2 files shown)


lupa/_lupa.pyx
```diff
@@ -1294,7 +1294,11 @@ cdef int raise_lua_error(LuaRuntime runtime, lua_State* L, int result) except -1
     elif result == lua.LUA_ERRMEM:
         raise MemoryError()
     else:
-        raise LuaError( build_lua_error_message(runtime, L, None, -1) )
+        error = py_from_lua(runtime, L, 1)
+        if isinstance(error, BaseException):
```
scoder, line 1298: Hmm. I'm not sure it's intentional that this retrieves a Python exception object. Couldn't such an exception also end up on the Lua stack for other reasons? Meaning, could this run into false positives?

synodriver:
I've also meet the same problem, and I think the extra check should be at ```resume_lua_thread```, right before ```raise_lua_error(thread._runtime, co, status)```, which only influence ```co.send``` itself.

PR #153, opened 2021-04-02
Add Windows binary file extensions to gitignore

guidanoli:


scoder:
That seems a bit much. You probably only wanted to add a single commit or so, right?

guidanoli:
> That seems a bit much. You probably only wanted to add a single commit or so, right?

You could squash all of the commits into one, right? It's just that I couldn't find a better way to isolate each feature...

PR #155, opened 2021-04-02
Safer typing in lock.pxi

guidanoli:
We should not use `-1` as a sentinel for "no owner", because the type Python uses for threads has changed from signed to unsigned, and is still a valid value for thread identification. What I did is type this field conditionally depending on the Python version (`unsigned long` with Python >= 3.7, `long` otherwise), and add the `has_owner` boolean field to avoid using unsafe sentinel values. Now only if `_has_owner` is `True`, `_owner` has a valid value.


lupa/_lupa.pyx
```diff
@@ -1182,13 +1188,21 @@ cdef object py_from_lua(LuaRuntime runtime, lua_State *L, int n):
...
         else:
             number = lua.lua_tonumber(L, n)
-            if number == <lua.lua_Integer>number:
-                return <lua.lua_Integer>number
+            integer = <lua.lua_Integer>number
+            if number == integer:
+                if maxint and integer >= -maxint and integer < maxint:
+                        return <int>integer
+                else:
+                        return integer
```
scoder: This seems worth a helper function. Also, the cast must be `<long>`, not `<int>`.

lupa/_lupa.pyx
```diff
@@ -50,6 +50,11 @@ try:
 except ImportError:
     from collections import Mapping  # Py2
 
+try:
+    from sys import maxint
+except ImportError:
+    maxint = None
```
scoder: `cdef long maxint`, then set `maxint = 0` instead of `None` in Py3.

lupa/lock.pxi
```diff
@@ -12,13 +23,14 @@ cdef class FastRLock:
...
-    cdef long _owner            # ID of thread owning the lock
+    cdef bint _has_owner        # whether the lock has an owner
+    cdef pythread_t _owner      # ID of thread owning the lock
     cdef int _count             # re-entry count
     cdef int _pending_requests  # number of pending requests for real lock
     cdef bint _is_locked        # whether the real lock is acquired
 
     def __cinit__(self):
-        self._owner = -1
+        self._has_owner = False
```
scoder, line 33: Use `<pythread_t> -1` as sentinel, maybe?

lupa/_lupa.pyx
guidanoli: That's not present in the current version of _lupa.pyx. You can safely ignore this and squash all commits into one.

lupa/_lupa.pyx
guidanoli: That's not present in the current version of _lupa.pyx. You can safely ignore this and squash all commits into one.

lupa/lock.pxi
guidanoli: Not really safe, because it's still a valid thread id.

lupa/_lupa.pyx
scoder: I'm not sure I understand. Do you want me to review a dozen separate features, arbitrarily split across 54 commits, without looking at separate commits and commit chains?

PR #165, opened 2021-04-02
Safer typing for Python thread ids

### LINKED ISSUE #179: Unsafe thread typing
**Issue:**
`FastRLock._owner` is typed as `long` but `PyThread_get_thread_ident` returns `unsigned long` for versions of Python 3.7.0a2 and later. See the discussion in [bpo-6532](https://bugs.python.org/issue6532), which led to this patch. It caused problems in logging signed thread IDs (as you may know, the size of `long` varies from platform to platform, so "signedness" shows itself at different limits).

**Solution:**
Both `pthread_t` (Linux) and `DWORD` (Windows) are unsigned, so it's a natural solution to make thread identifiers in Python unsigned too. Unfortunately, we need to keep compatibility with versions prior to Python 3.7.0a2. So the best way is to conditionally define a `pythead_t` depending on the current version of Python.

Moreover, we should not use `-1` as sentinel for `FastRLock._owner` because there is no guarantee a thread cannot have this id (in some system). We should then rely on `FastRLock._count`, which is `0` when not owned.

Furthermore, counters should be unsigned for obvious reasons.

guidanoli:
* Since Python 3.7, thread ids are unsigned. Before, they were signed.
So we cannot rely on storing '-1' as a sentinel value anymore. Thus, we
should rely on the `_count` field.
* Satisfies compiler complaints about unsafe type casts

Closes #179


lupa/lock.pxi
```diff
@@ -12,13 +23,14 @@ cdef class FastRLock:
     wonderful GIL.
     """
     cdef pythread.PyThread_type_lock _real_lock
-    cdef long _owner            # ID of thread owning the lock
+    cdef bint _has_owner        # whether the lock has an owner
```
scoder, line 26: This is not needed. You can see from `_count > 0` that the lock is owned.

guidanoli: Fixed in https://github.com/scoder/lupa/pull/165/commits/4c3012a9308bcf5060132ebdce001ed60bf6b2a0

guidanoli:
@scoder another motivation for this change is to silence the compiler about unsafe type casts:
```
/home/guilherme/guidanoli/lupa-from-lua/lupa/lupa/_lupa.c: In function ‘__pyx_pf_4lupa_5_lupa_9FastRLock_6release’:
/home/guilherme/guidanoli/lupa-from-lua/lupa/lupa/_lupa.c:3194:38: warning: comparison of integer expressions of different signedness: ‘long int’ and ‘long unsigned int’ [-Wsign-compare]
   __pyx_t_1 = ((__pyx_v_self->_owner != PyThread_get_thread_ident()) != 0);
                                      ^~
/home/guilherme/guidanoli/lupa-from-lua/lupa/lupa/_lupa.c: In function ‘__pyx_pf_4lupa_5_lupa_9FastRLock_10__exit__’:
/home/guilherme/guidanoli/lupa-from-lua/lupa/lupa/_lupa.c:3416:38: warning: comparison of integer expressions of different signedness: ‘long int’ and ‘long unsigned int’ [-Wsign-compare]
   __pyx_t_1 = ((__pyx_v_self->_owner != PyThread_get_thread_ident()) != 0);
                                      ^~
/home/guilherme/guidanoli/lupa-from-lua/lupa/lupa/_lupa.c: In function ‘__pyx_pf_4lupa_5_lupa_9FastRLock_12_is_owned’:
/home/guilherme/guidanoli/lupa-from-lua/lupa/lupa/_lupa.c:3511:59: warning: comparison of integer expressions of different signedness: ‘long int’ and ‘long unsigned int’ [-Wsign-compare]
   __pyx_t_1 = __Pyx_PyBool_FromLong((__pyx_v_self->_owner == PyThread_get_thread_ident())); if (unlikely(!__pyx_t_1)) __PYX_ERR(1, 55, __pyx_L1_error)
```

scoder:
Thanks!

PR #167, opened 2021-04-03
Allow for building Lupa with custom Lua library

guidanoli:
* Adds the following command-line arguments to setup.py:
  * `--lua-lib <lib-file>` for the Lua library file path, e.g. `/usr/local/bin/lua.a`
  * `--lua-includes <inc-dir>` for the Lua library header directory path, e.g. `/usr/local/bin/include`
* Adds Lua 5.4 to the list of Lua version to look for in the system
* Adds documentation about the usage of these setup options
* Replaced `os_path` by `os.path` to improve code readability at the
expense of negligible performance gain

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -927,3 +927,23 @@ following should work on a Linux system:
       >>> posix_module = lua.require('posix')     # doctest: +SKIP
 
 .. _luaposix: http://git.alpinelinux.org/cgit/luaposix
+
+Building with different Lua versions
+------------------------------------
+
+If you wish to build Lupa with a specific version of Lua, you can
+configure the following options on setup:
+
+.. list-table::
+   :widths: 20 20 20
+   :header-rows: 1
+   
+   * - Option
+     - Description
+     - Example
+   * - ``--lua-lib <libfile>``
+     - Lua library file path
+     - ``--lua-lib /usr/local/lib/lualib.a``
+   * - ``--lua-includes <incdir>``
+     - Lua include directory
+     - ``--lua-includes /usr/local/include``
--- a/README.rst
+++ b/README.rst
@@ -947,3 +947,12 @@ configure the following options on setup:
    * - ``--lua-includes <incdir>``
      - Lua include directory
      - ``--lua-includes /usr/local/include``
+   * - ``--use-bundle``
+     - Use bundled Lua
+     -
+   * - ``--no-bundle``
+     - Don't use bundled Lua
+     -
+   * - ``--no-lua-jit``
+     - Don't use LuaJIT
+     -
--- a/setup.py
+++ b/setup.py
@@ -215,7 +215,7 @@ def use_bundled_lua(path, lua_sources, macros):
 
 
 def get_option(name):
-    for i, arg in enumerate(sys.argv[[1:], 1):
+    for i, arg in enumerate(sys.argv[1:-1], 1):
         if arg == name:
             sys.argv.pop(i)
             return sys.argv.pop(i)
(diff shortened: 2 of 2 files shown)


setup.py
```diff
@@ -197,6 +214,14 @@ def use_bundled_lua(path, lua_sources, macros):
...
 
 
+def get_option(name):
+    for i, arg in enumerate(sys.argv):
+        if i == 0:
+            continue # Ignore script name
+        if arg == name and i < len(sys.argv) - 1:
+            sys.argv.pop(i)
+            return sys.argv.pop(i)
+
```
scoder, line 224:
```suggestion
def get_option(name):
    for i, arg in enumerate(sys.argv[[1:], 1):
        if arg == name:
            sys.argv.pop(i)
            return sys.argv.pop(i)
    return ""


```

setup.py
```diff
@@ -115,25 +115,42 @@ def lua_libs(package='luajit'):
...
+    print('Using Lua include directory: %s' % lua_I)
+
+    root, ext = os.path.splitext(lua_l)
+    if os.name == 'nt' and ext == '.lib':
+        return dict(extra_objects=[lua_l],
+                    include_dirs=[lua_I],
+                    libfile=lua_l)
+
+    return dict(extra_objects=[lua_l],
+                include_dirs=[lua_I])
```
scoder, line 135:
```suggestion
    else:
        return dict(extra_objects=[lua_l],
                    include_dirs=[lua_I])
```

guidanoli:
@scoder How about now?

scoder:
Should have added a comment, I guess. I meant to include the last item in the slice, because I would like to let `setup.py --lua-L` raise an exception rather than passing silently.

setup.py
```diff
@@ -115,25 +115,42 @@ def lua_libs(package='luajit'):
 
 basedir = os.path.abspath(os.path.dirname(__file__))
 
+def get_lua_build_from_arguments():
+    lua_l = get_option('--lua-l')
+    lua_I = get_option('--lua-I')
```
scoder, line 120: These are difficult to distinguish in some fonts. Let's make the names explicit.
```python
    lua_lib = get_option('--lua-lib')
    lua_includes = get_option('--lua-includes')
```

guidanoli: Resolved in 32f0d3c

scoder:
Ok, I can live with that. Since it adds yet another way of configuring the build, would you mind adding a documentation section at the end of `README.rst` about the different options? Maybe "Building with different Lua versions"? They don't even show up in `setup.py --help`, so users will have to read and understand the setup script in order to find out how to make it use the right Lua version on their side.

guidanoli:
> Ok, I can live with that. Since it adds yet another way of configuring the build, would you mind adding a documentation section at the end of `README.rst` about the different options? Maybe "Building with different Lua versions"? They don't even show up in `setup.py --help`, so users will have to read and understand the setup script in order to find out how to make it use the right Lua version on their side.

Sure

README.rst
```diff
@@ -927,3 +927,23 @@ following should work on a Linux system:
...
+   
+   * - Option
+     - Description
+     - Example
+   * - ``--lua-lib <libfile>``
+     - Lua library file path
+     - ``--lua-lib /usr/local/lib/lualib.a``
+   * - ``--lua-includes <incdir>``
+     - Lua include directory
+     - ``--lua-includes /usr/local/include``
```
scoder, line 949: … plus all the other options, like `--use-bundle`, `pkg-config`, etc. :)

guidanoli: Ok!

scoder:
Ok, let's get this in.

scoder:
Thanks!

guidanoli:
> Thanks!

You're welcome!

Issue #146, opened 2020-08-17
Add integer support (Lua 5.3)


scoder:
Lua 5.3 has [introduced](https://www.lua.org/manual/5.3/readme.html#changes) support for integer values, in addition to the previous C double floating point values. Lupa should support these and map Python `int` objects to Lua integers, it possible. It seems ok to drop support for Lua 5.1/2 for this feature, then releasing as Lupa 2.0.

- add tests
- add the new type to `py_from_lua()` and `py_to_lua()`.
- find a reasonable fallback for large integers (preferably configurable as `LuaRuntime` parameter), either rejecting them or converting them to a Lua float.

mfrigerio17:
Is this solved in Lupa 1.10?
I still get a floating point number in Lua when using Python's `len` builtin:

```python
import lupa
lua = lupa.LuaRuntime()
f = lua.execute("return function(seq) return print( python.builtins.len(seq) ), print(4) end")
f( ['a', 'b', 'c', 'd'] )
```
This prints:
```
4.0
4
```
I expect the integer `4` also for the first print.

Python 3.9.5
Lua 5.3.3
Lupa 1.10

If I _return_ the length instead of printing it (within Lua), the return value in Python is actually an integer:

```python
f = lua.execute("return function(seq) return python.builtins.len(seq) end")
type( f( ['a', 'b', 'c', 'd'] ) )
```
Prints: `int`

scoder:
Yes, it's in 1.10. Can't say right now what goes wrong here. Could you investigate a bit deeper?

Satellite-Galaxy:
I am not sure this was fixed I am using 1.13 and am having the same issue.

PR #168, opened 2021-04-03
Adds support for Lua 5.3 integers 

guidanoli:
* Lua -> Python
    * For Lua >= 5.3
      * If Lua number is of subtype integer, converts to Python int
      * If Lua number is of subtype float, converts to Python float
    * For Lua < 5.3 (current algorithm)
      * If Lua number has decimal part, converts to Python float
      * If Lua number doesn't have decimal part, converts to Python int
* Python -> Lua
    * For Python 2.7
      * If number is int, converts to Lua integer
    * For every Python
      * If number is long, converts to Lua integer
    * If overflow occurrs
      * If the global function `lupa_overflow_cb` exists, it is called with the Python object, expecting the returned value to be the "treated object" (e.g. converted to float)
      * If the global function `lupa_overflow_cb` doesn't exist or if it fails, an error is raised

Addresses #146.


lupa/lua.pxd
```diff
@@ -439,14 +439,15 @@ cdef extern from *:
     #define lua_objlen(L, i) lua_rawlen(L, (i))
     #endif
 
-    #if LUA_VERSION_NUM >= 504
-    #define read_lua_version(L)  ((int) lua_version(L))
-    #elif LUA_VERSION_NUM >= 502
-    #define read_lua_version(L)  ((int) *lua_version(L))
-    #elif LUA_VERSION_NUM >= 501
```
scoder, line 446: Why did you remove this? It is meant to read the runtime version of Lua that is used, not the compile-time version.

lupa/_lupa.pyx
```diff
@@ -1222,9 +1256,6 @@ cdef int py_to_lua(LuaRuntime runtime, lua_State *L, object o, bint wrap_none=Fa
             raise LuaError("cannot mix objects from different Lua runtimes")
         lua.lua_rawgeti(L, lua.LUA_REGISTRYINDEX, (<_LuaObject>o)._ref)
         pushed_values_count = 1
-    elif isinstance(o, float):
-        lua.lua_pushnumber(L, <lua.lua_Number><double>o)
-        pushed_values_count = 1
```
scoder, line 1227: This is needed for subclasses of `float`.

lupa/_lupa.pyx
```diff
@@ -1207,11 +1235,17 @@ cdef int py_to_lua(LuaRuntime runtime, lua_State *L, object o, bint wrap_none=Fa
         lua.lua_pushnumber(L, <lua.lua_Number>cpython.float.PyFloat_AS_DOUBLE(o))
         pushed_values_count = 1
     elif isinstance(o, long):
-        lua.lua_pushnumber(L, <lua.lua_Number>cpython.long.PyLong_AsDouble(o))
-        pushed_values_count = 1
+        try:
+            lua.lua_pushinteger(L, <lua.lua_Integer>cpython.long.PyLong_AsLongLong(o))
```
scoder, line 1239: This is not safe if `sizeof(lua_Integer)` < `sizeof(long long)`.
```suggestion
            lua.lua_pushinteger(L, <lua.lua_Integer>o)
```

lupa/_lupa.pyx
```diff
@@ -1207,11 +1235,17 @@ cdef int py_to_lua(LuaRuntime runtime, lua_State *L, object o, bint wrap_none=Fa
...
+            pushed_values_count = 1
+        except Exception as e:
```
scoder, line 1241:
```suggestion
        except OverflowError as e:
```

lupa/_lupa.pyx
```diff
@@ -1207,11 +1235,17 @@ cdef int py_to_lua(LuaRuntime runtime, lua_State *L, object o, bint wrap_none=Fa
...
+            overflow_exception = e
     elif IS_PY2 and isinstance(o, int):
-        lua.lua_pushnumber(L, <lua.lua_Number><long>o)
-        pushed_values_count = 1
+        try:
+            lua.lua_pushinteger(L, <lua.lua_Integer><long>o)
+            pushed_values_count = 1
+        except Exception as e:
```
scoder, line 1247:
```suggestion
            lua.lua_pushinteger(L, <lua.lua_Integer>o)
            pushed_values_count = 1
        except OverflowError as e:
```

lupa/_lupa.pyx
```diff
@@ -1183,9 +1191,29 @@ cdef int py_function_result_to_lua(LuaRuntime runtime, lua_State *L, object o) e
          return len(<tuple>o)
      return py_to_lua(runtime, L, o)
 
+cdef int py_to_lua_overflow(LuaRuntime runtime, lua_State *L, object o) except -1:
```
scoder, line 1194: Maybe `py_to_lua_handle_overflow` ?

lupa/_lupa.pyx
```diff
@@ -1183,9 +1191,29 @@ cdef int py_function_result_to_lua(LuaRuntime runtime, lua_State *L, object o) e
...
+    cdef const char* cb_name = "lupa_overflow_cb"
```
scoder, line 1195: This needs to be documented and tested. In fact, I wonder if this should be a Lua-level or Python-level feature. Or both.

scoder (commented):
I think this is going in the right direction. Tests are missing. And it adds a new callback function which needs to be properly tested and documented.

lupa/lua.pxd
guidanoli: You're right. This is not necessary to change this for this specific PR.

lupa/_lupa.pyx
guidanoli: You're right. I thought this was a mistake.

lupa/_lupa.pyx
guidanoli: The overflow only occurrs when converting a Python int/long to Lua integer, so I think it should be manipulated in Lua since the value needs to be of a Lua type.

guidanoli: Ok, I'll be documenting and allowing for this callback to be set from both sides (in Lua and in Python). Also, adding tests.

lupa/_lupa.pyx
```diff
@@ -423,6 +435,13 @@ cdef class LuaRuntime:
             lua.lua_settop(L, old_top)
             unlock_runtime(self)
 
+    def set_overflow_handler(self, overflow_handler):
+        """Set overflow handler.
```
scoder, line 439:
```suggestion
        """Set the overflow handler function that is called on failures to pass large numbers to Lua.
```

lupa/_lupa.pyx
```diff
@@ -1212,11 +1218,11 @@ cdef int py_function_result_to_lua(LuaRuntime runtime, lua_State *L, object o) e
      return py_to_lua(runtime, L, o)
 
 cdef int py_to_lua_handle_overflow(LuaRuntime runtime, lua_State *L, object o) except -1:
-    cdef const char* cb_name = "lupa_overflow_cb"
-    cdef int nargs = 0
+    cdef int nargs
 
-    lua.lua_getglobal(L, cb_name)
-    if lua.lua_isfunction(L, -1):
+    lua.lua_pushlstring(L, "LUPAOFH", 7)
```
scoder, line 1223: Unless we can avoid adding this global name entirely, let's at least make it long and explicit, to reduce the chances of a naming collision in Lua code.
For example: `_LUPA_NUMBER_OVERFLOW_CALLBACK_FUNCTION`. And please put it into a global constant, like `POBJECT` further up.

guidanoli: It's actually in the register, not in the global scope!

guidanoli: Resolved in 6fb99e6

lupa/_lupa.pyx
guidanoli: Resolved in https://github.com/scoder/lupa/pull/168/commits/f112b9b6ba48622a71c5fd50c1f34e43fdec9e46

lupa/_lupa.pyx
```diff
@@ -423,17 +435,33 @@ cdef class LuaRuntime:
...
 
+    def set_overflow_handler(self, overflow_handler):
+        """Set the overflow handler function that is called on failures to pass large numbers to Lua.
+        """
+        cdef lua_State *L = self._state
+
+        if overflow_handler is not None and not callable(overflow_handler):
+            raise ValueError("overflow_handler must be callable")
+        
+        lua.lua_pushlstring(L, LUPAOFH, LUPAOFH_LEN)
```
scoder, line 446: There is no performance issue here.
```suggestion
        lua.lua_pushlstring(L, LUPAOFH, len(LUPAOFH))
```

lupa/_lupa.pyx
```diff
@@ -63,6 +63,8 @@ except ImportError:
     import builtins
 
 DEF POBJECT = b"POBJECT" # as used by LunaticPython
+DEF LUPAOFH = b"LUPA_NUMBER_OVERFLOW_CALLBACK_FUNCTION"
+DEF LUPAOFH_LEN = len(LUPAOFH)
```
scoder, line 67:
```suggestion
```

lupa/_lupa.pyx
```diff
@@ -1183,9 +1219,29 @@ cdef int py_function_result_to_lua(LuaRuntime runtime, lua_State *L, object o) e
          return len(<tuple>o)
      return py_to_lua(runtime, L, o)
 
+cdef int py_to_lua_handle_overflow(LuaRuntime runtime, lua_State *L, object o) except -1:
+    cdef int nargs
+
+    lua.lua_pushlstring(L, LUPAOFH, LUPAOFH_LEN)
```
scoder, line 1225:
```suggestion
    lua.lua_pushlstring(L, LUPAOFH, len(LUPAOFH))
```

lupa/_lupa.pyx
```diff
@@ -1207,11 +1263,17 @@ cdef int py_to_lua(LuaRuntime runtime, lua_State *L, object o, bint wrap_none=Fa
         lua.lua_pushnumber(L, <lua.lua_Number>cpython.float.PyFloat_AS_DOUBLE(o))
         pushed_values_count = 1
     elif isinstance(o, long):
-        lua.lua_pushnumber(L, <lua.lua_Number>cpython.long.PyLong_AsDouble(o))
-        pushed_values_count = 1
+        try:
+            lua.lua_pushinteger(L, <lua.lua_Integer>o)
+            pushed_values_count = 1
+        except OverflowError as e:
+            overflow_error = e
```
scoder, line 1270: I think it would be nice if we could allow users to look at the exception in `sys.exc_value`.
```suggestion
        except OverflowError:
            pushed_values_count = py_to_lua_handle_overflow(runtime, L, o)
            if pushed_values_count <= 0:
                raise
```

lupa/_lupa.pyx
```diff
@@ -1207,11 +1263,17 @@ cdef int py_to_lua(LuaRuntime runtime, lua_State *L, object o, bint wrap_none=Fa
...
     elif IS_PY2 and isinstance(o, int):
-        lua.lua_pushnumber(L, <lua.lua_Number><long>o)
-        pushed_values_count = 1
+        try:
+            lua.lua_pushinteger(L, <lua.lua_Integer>o)
+            pushed_values_count = 1
+        except OverflowError as e:
+            overflow_error = e
```
scoder, line 1276:
```suggestion
        except OverflowError:
            pushed_values_count = py_to_lua_handle_overflow(runtime, L, o)
            if pushed_values_count <= 0:
                raise
```

lupa/_lupa.pyx
```diff
@@ -1183,9 +1219,29 @@ cdef int py_function_result_to_lua(LuaRuntime runtime, lua_State *L, object o) e
...
+        else:
+            return 1
+    else:
+        lua.lua_pop(L, 1)
+        return 0
+
 cdef int py_to_lua(LuaRuntime runtime, lua_State *L, object o, bint wrap_none=False) except -1:
     cdef int pushed_values_count = 0
     cdef int type_flags = 0
+    cdef object overflow_error = None
```
scoder, line 1244:
```suggestion
```

lupa/_lupa.pyx
```diff
@@ -1233,6 +1295,12 @@ cdef int py_to_lua(LuaRuntime runtime, lua_State *L, object o, bint wrap_none=Fa
             # prefer __getitem__ over __getattr__ by default
             type_flags = OBJ_AS_INDEX if hasattr(o, '__getitem__') else 0
         pushed_values_count = py_to_lua_custom(runtime, L, o, type_flags)
+
+    if overflow_error is not None:
+        pushed_values_count = py_to_lua_handle_overflow(runtime, L, o)
+        if pushed_values_count <= 0:
+            raise overflow_error
+
```
scoder, line 1303:
```suggestion
```

lupa/tests/test.py
```diff
@@ -2619,6 +2619,119 @@ def test_nil_debug_traceback(self):
...
             self.assertNotIn("stack traceback:", e.args[0])
 
+################################################################################
+# tests for handling overflow
+
+class TestOverflow(SetupLuaRuntimeMixin):
+    maxinteger = sys.maxsize # maximum value for C Py_ssize_t
+    biginteger = (maxinteger + 1) << 1 # value too big to fit in C size_t
+    maxfloat = sys.float_info.max # maximum value for Python float
+    bigfloat = int(maxfloat) * 2 # value too big to fit in Python float
```
scoder, line 2629: PEP-8 wants two spaces before the comment, but for the sake of readability, let's just align them.
```suggestion
    maxinteger = sys.maxsize            # maximum value for C Py_ssize_t
    biginteger = (maxinteger + 1) << 1  # value too big to fit in C size_t
    maxfloat = sys.float_info.max       # maximum value for Python float
    bigfloat = int(maxfloat) * 2        # value too big to fit in Python float
```

lupa/tests/test.py
```diff
@@ -2619,6 +2619,119 @@ def test_nil_debug_traceback(self):
         except lupa.LuaError as e:
             self.assertNotIn("stack traceback:", e.args[0])
```
scoder, line 2621: PEP-8 wants two empty lines between top-level classes. Also below, please.
```suggestion


```

lupa/tests/test.py
```diff
@@ -2619,6 +2619,119 @@ def test_nil_debug_traceback(self):
...
+
+class TestOverflowWithObjectHandler(TestOverflow, unittest.TestCase):
+
+    def test_overflow(self):
+        self.lua.execute('python.set_overflow_handler(function(o) return o end)')
+        self.assertEqual(self.lua.eval('type')(int(self.maxfloat)), 'userdata')
+
+class TestBadOverflowHandlerInPython(unittest.TestCase):
+    def test_error(self):
+        self.assertRaises(ValueError, lupa.LuaRuntime, overflow_handler=123)
```
scoder, line 2687: This suggests to me that it should in fact raise a `TypeError`.

lupa/tests/test.py
```diff
@@ -2619,6 +2619,119 @@ def test_nil_debug_traceback(self):
...
+        self._test_set_overflow_handler('false')
+
+    def test_string(self):
+        self._test_set_overflow_handler('"abc"')
+
+    def test_thread(self):
+        self._test_set_overflow_handler('coroutine.create(function() end)')
+
+    def _test_set_overflow_handler(self, overflow_handler_code):
+        self.assertRaises(lupa.LuaError, self.lua.execute, 'python.set_overflow_handler(%s)' % overflow_handler_code)
```
scoder, line 2710: I would move these main test 'helper' functions up to the top. Classes are usually read from top to bottom, so this would help in understanding what is going on in this class.

scoder (commented):
Mostly style issues.
Note that you can collect multiple suggestions in one commit batch.

lupa/_lupa.pyx
```diff
@@ -1799,15 +1863,26 @@ cdef int py_iter_next_with_gil(lua_State* L, py_object* py_iter) with gil:
...
         finally: return -1
 
+# overflow handler setter
+
+cdef int py_set_overflow_handler(lua_State* L) nogil:
+    if not lua.lua_isnil(L, 1) and not lua.lua_isfunction(L, 1):
+        return lua.luaL_argerror(L, 1, "expected nil or function")
+    lua.lua_pushlstring(L, LUPAOFH, LUPAOFH_LEN) # hdl ... name
+    lua.lua_pushvalue(L, 1)                      # hdl ... name hdl
+    lua.lua_rawset(L, lua.LUA_REGISTRYINDEX)     # hdl ...
```
scoder, line 1873:
```suggestion
    lua.lua_pushlstring(L, LUPAOFH, len(LUPAOFH))  # hdl ... name
    lua.lua_pushvalue(L, 1)                        # hdl ... name hdl
    lua.lua_rawset(L, lua.LUA_REGISTRYINDEX)       # hdl ...
```

lupa/tests/test.py
guidanoli: But I do the same checking as for what is done with `attribute_filter`. I just check if it is `None` or callable.

lupa/_lupa.pyx
```diff
@@ -1868,9 +1868,9 @@ cdef int py_iter_next_with_gil(lua_State* L, py_object* py_iter) with gil:
 cdef int py_set_overflow_handler(lua_State* L) nogil:
     if not lua.lua_isnil(L, 1) and not lua.lua_isfunction(L, 1):
         return lua.luaL_argerror(L, 1, "expected nil or function")
-    lua.lua_pushlstring(L, LUPAOFH, LUPAOFH_LEN) # hdl ... name
-    lua.lua_pushvalue(L, 1)                      # hdl ... name hdl
-    lua.lua_rawset(L, lua.LUA_REGISTRYINDEX)     # hdl ...
+    lua.lua_pushlstring(L, LUPAOFH, len(LUPAOFH))  # hdl ... name
```
guidanoli, line 1871: This raises the following error:
`Calling gil-requiring function not allowed without gil`
I've opened an issue about the matter (https://github.com/cython/cython/issues/4093)
For the mean time, there is really no other way but to `DEF`-ine `len(LUPAOFH)` elsewhere.

lupa/tests/test.py
guidanoli: Resolved in https://github.com/scoder/lupa/pull/168/commits/1f36c0f05f803a3eee06b316c599d84c7625f372

lupa/tests/test.py
scoder: Hmm. I guess that was a bad decision back then as well. Not sure if it matters enough to introduce an inconsistency, though. Let's stick with what's there.

guidanoli:
@scoder Is this pull request missing anything?

lupa/_lupa.pyx
```diff
@@ -1207,11 +1261,21 @@ cdef int py_to_lua(LuaRuntime runtime, lua_State *L, object o, bint wrap_none=Fa
...
     elif IS_PY2 and isinstance(o, int):
-        lua.lua_pushnumber(L, <lua.lua_Number><long>o)
-        pushed_values_count = 1
+        try:
+            lua.lua_pushinteger(L, <lua.lua_Integer>o)
+            pushed_values_count = 1
+        except OverflowError:
+            pushed_values_count = py_to_lua_handle_overflow(runtime, L, o)
+            if pushed_values_count <= 0:
+                raise
```
scoder, line 1278: This is entirely the same implementation now, right? Just replace the condition with `elif isinstance(o, (long, int)):`.

guidanoli:
@scoder should be good to go

lupa/_lupa.pyx
```diff
@@ -1895,14 +1888,14 @@ cdef int py_iter_next_with_gil(lua_State* L, py_object* py_iter) with gil:
...
 cdef int py_set_overflow_handler(lua_State* L) nogil:
-    if not lua.lua_isnil(L, 1) and not lua.lua_isfunction(L, 1):
-        return lua.luaL_argerror(L, 1, "expected nil or function")
-    lua.lua_pushlstring(L, LUPAOFH, LUPAOFH_LEN)   # hdl ... name
-    lua.lua_pushvalue(L, 1)                        # hdl ... name hdl
-    lua.lua_rawset(L, lua.LUA_REGISTRYINDEX)       # hdl ...
+    if not lua.lua_isnil(L, 1) and \
+       not lua.lua_isfunction(L, 1) and \
+       not unpack_python_argument_or_jump(L, 1):
+        return lua.luaL_argerror(L, 1, "expected nil, function or Python object")
```
scoder, line 1895:
```suggestion
    if (not lua.lua_isnil(L, 1)
            and not lua.lua_isfunction(L, 1)
            and not unpack_python_argument_or_jump(L, 1)):
        return lua.luaL_argerror(L, 1, "expected nil, function or Python object")
```

lupa/tests/test.py
```diff
@@ -2697,9 +2703,6 @@ def test_number(self):
     def test_table(self):
         self._test_set_overflow_handler('{}')
 
-    def test_userdata(self):
-        self._test_set_overflow_handler('python.builtins.float')
-
```
scoder: Why did this have to go?

lupa/_lupa.pyx
```diff
@@ -1191,6 +1226,25 @@ cdef int py_function_result_to_lua(LuaRuntime runtime, lua_State *L, object o) e
...
+            lua.lua_pop(L, 1)
+            return 0
+        if lua.lua_pcall(L, nargs, 1, 0):
+            lua.lua_pop(L, 1)
+            return 0
+        else:
+            return 1
+    else:
+        lua.lua_pop(L, 1)
+        return 0
```
scoder, line 1246: Let's handle all error cases the same.
```suggestion
    if lua.lua_isnil(L, -1):
        lua.lua_pop(L, 1)
        return 0
    nargs = py_to_lua_custom(runtime, L, o, 0)
    if nargs <= 0:
        lua.lua_pop(L, 1)
        return 0
    if lua.lua_pcall(L, nargs, 1, 0):
        lua.lua_pop(L, 1)
        return 0
    return 1
```

lupa/tests/test.py
```diff
@@ -2620,11 +2620,127 @@ def test_nil_debug_traceback(self):
...
+        self.assertMathType(3.14, 'float')
+        self.assertMathType(-3.14, 'float')
+        self.assertMathType(self.maxfloat, 'float')
+        self.assertMathType(-self.maxfloat, 'float')
+
+    def assertMathType(self, number, math_type):
+        self.assertEqual(self.lua_type(number), 'number')
+        if self.lua_math_type is not None:
+            self.assertEqual(self.lua_math_type(number), math_type)
+
```
scoder, line 2663:
```suggestion


```

lupa/tests/test.py
```diff
@@ -2620,11 +2620,127 @@ def test_nil_debug_traceback(self):
...
+class TestOverflowWithoutHandler(TestOverflowMixin, unittest.TestCase):
+    lua_runtime_kwargs = dict(overflow_handler=None)
+
+    def test_overflow(self):
+        self.assertRaises(OverflowError, self.assertMathType, self.biginteger, 'integer')
+        self.assertRaises(OverflowError, self.assertMathType, int(self.maxfloat), 'integer')
+        self.assertRaises(OverflowError, self.assertMathType, self.bigfloat, 'integer')
+
```
scoder, line 2671:
```suggestion


```

lupa/tests/test.py
```diff
@@ -2620,11 +2620,127 @@ def test_nil_debug_traceback(self):
...
+class TestOverflowWithFloatHandler(TestOverflowMixin, unittest.TestCase):
+    lua_runtime_kwargs = dict(overflow_handler=float)
+
+    def test_overflow(self):
+        self.assertMathType(self.biginteger, 'float')
+        self.assertMathType(int(self.maxfloat), 'float')
+        self.assertRaises(OverflowError, self.assertMathType, self.bigfloat, 'float')
+
```
scoder, line 2679:
```suggestion


```

lupa/tests/test.py
```diff
@@ -2620,11 +2620,127 @@ def test_nil_debug_traceback(self):
...
+class TestOverflowWithObjectHandler(TestOverflowMixin, unittest.TestCase):
+    def test_overflow(self):
+        self.lua.execute('python.set_overflow_handler(function(o) return o end)')
+        self.assertEqual(self.lua.eval('type')(int(self.maxfloat)), 'userdata')
+
```
scoder, line 2684:
```suggestion


```

lupa/tests/test.py
```diff
@@ -2620,11 +2620,127 @@ def test_nil_debug_traceback(self):
...
+class TestFloatOverflowHandlerInLua(TestOverflowMixin, unittest.TestCase):
+    def test_overflow(self):
+        self.lua.execute('python.set_overflow_handler(python.builtins.float)')
+        self.assertMathType(self.biginteger, 'float')
+        self.assertMathType(int(self.maxfloat), 'float')
+        self.assertRaises(OverflowError, self.assertMathType, self.bigfloat, 'float')
+
```
scoder, line 2691:
```suggestion


```

lupa/tests/test.py
```diff
@@ -2620,11 +2620,127 @@ def test_nil_debug_traceback(self):
...
+class TestBadOverflowHandlerInPython(unittest.TestCase):
+    def test_error(self):
+        self.assertRaises(ValueError, lupa.LuaRuntime, overflow_handler=123)
+
```
scoder, line 2695:
```suggestion


```

lupa/tests/test.py
```diff
@@ -2620,11 +2620,127 @@ def test_nil_debug_traceback(self):
...
+    def test_boolean(self):
+        self._test_set_overflow_handler('true')
+        self._test_set_overflow_handler('false')
+
+    def test_string(self):
+        self._test_set_overflow_handler('"abc"')
+
+    def test_thread(self):
+        self._test_set_overflow_handler('coroutine.create(function() end)')
+
```
scoder, line 2715:
```suggestion


```

scoder:
Very nice, thanks!

lupa/tests/test.py
guidanoli: Because we now accept userdata as overflow handlers.

guidanoli: We could test not POBJECT userdata in earlier versions Lua where newproxy is available, but not on the latest versions.

guidanoli:
> Very nice, thanks!

My pleasure!

scoder:
Turns out, this implementation has 32bit issues on Windows, e.g.
https://ci.appveyor.com/project/St/lupa/build/job/m82mwnlnj0k5p6py
Lua and Python apparently seem to disagree about the integer size.
I already fixed the implementation to return (Python) `int` in Py2 instead of always returning (Python) `long` there, but there are still issues with the tests (and probably only the tests).

I think we should expose the range of `lua_Integer` in the API. That would solve this problem, and would allow users to take better actions.

guidanoli:
> I think we should expose the value range of `lua_Integer` in the API. That would solve this problem, and would allow users to take better actions. Maybe, `lupa.MAX_LUA_INTEGER` and `lupa.MIN_LUA_INTEGER` ?

In what case exposing these constants would help users?

scoder:
> > I think we should expose the value range of `lua_Integer` in the API. That would solve this problem, and would allow users to take better actions. Maybe, `lupa.MAX_LUA_INTEGER` and `lupa.MIN_LUA_INTEGER` ?
> 
> In what case would these constants help users?

It's not obvious for a user of Lua, Lupa or Python what the supported integer range is on a given platform and Lua installation. Thus, providing these constants would allow users to understand why an overflow occurred and/or to take actions based on the kind of overflow that happened or might potentially happen for a given integer value. They might also be able to tell up-front that some calculation isn't possible on the current platform or could arrange for the specific value range in the choice of an algorithm or implementation.

Basically, if you can't hide it, make it transparent.

guidanoli:
I see, Python and Lua disagree on what is the biggest integer
Python 2: sys.maxint
Python 3: sys.maxsize
Lua (>= 5.3): math.maxsize
I'll be doing some tests in Windows x86 and releasing a new pull request soon.

guidanoli:
Experiments:
On Python3: ``python3 -c "import sys; print(sys.maxsize)"  -> 9223372036854775807``
On Python2: ``python2 -c "import sys; print(sys.maxint)" -> 2147483647``
On Lua 5.4: ``lua -e "print(math.maxinteger)"``  -> 9223372036854775807

PR #169, opened 2021-04-03
Add Lua>=5.3 integer support

SamblySok0:
with using lua.lua_tointeger if number equal to long number


scoder:
Thanks, but #168 seems more complete.

PR #171, opened 2021-04-05
Fixes Python missing references

guidanoli:
* When a Python object in Lua is ressurected, it may result in a missing
reference because its reference count has been zeroed and Python can't
ressurect it like Lua. This would lead to the dangling pointer problem.
The way to solve this is by tagging the Python object in Lua when it is
deallocated so that if it is later accessed, an error is raised.
* Optimized `LuaRuntime._pyrefs_in_lua`: Created the Cython class called
`_PyReference`, which contains the Python object being referenced and
the reference count (from Lua `userdata`)
* Added tests for missing references
* Added `lua_State*` parameter to `_LuaObject.push_lua_object` in order to
call `lua_rawgeti` with the correct Lua state, and substituted calls to
`lua_rawgeti` to `_LuaObject.push_lua_object` because it checks whether the
object is missing!
* Added `unpack_python_argument_or_jump` for checking whether the Python
object is missing and doing a long jump if so. This makes the code less
repetitive too.

### DIFF
--- a/lupa/_lupa.pyx
+++ b/lupa/_lupa.pyx
@@ -80,7 +80,6 @@ cdef struct py_object:
     PyObject* obj
     PyObject* runtime
     int type_flags  # or-ed set of WrappedObjectFlags
-    uintptr_t refcount
 
 
 include "lock.pxi"
--- a/lupa/_lupa.pyx
+++ b/lupa/_lupa.pyx
@@ -1256,39 +1256,45 @@ cdef bint py_to_lua_custom(LuaRuntime runtime, lua_State *L, object o, int type_
     cdef py_object* py_obj
     cdef object refkey = get_pyref_key(<PyObject*>o, type_flags)
     cdef _PyReference pyref
+
     lua.lua_getfield(L, lua.LUA_REGISTRYINDEX, PYREFST)  # tbl
+
+    # check if python object is already referenced in Lua
     if refkey in runtime._pyrefs_in_lua:
         pyref = <_PyReference>runtime._pyrefs_in_lua[refkey]
         lua.lua_rawgeti(L, -1, pyref._ref)              # tbl udata
         py_obj = <py_object*>lua.lua_touserdata(L, -1)
-        if not py_obj:
-            lua.lua_pop(L, 2)                           #
-            return 0 # values pushed
-        py_obj.ref_count += 1
-        lua.lua_remove(L, -2)                           # udata
-        return 1 # values pushed
-    else:
-        py_obj = <py_object*>lua.lua_newuserdata(L, sizeof(py_object))
-        if not py_obj:
-            lua.lua_pop(L, 1)                #
-            return 0 # values pushed
-        py_obj.obj = <PyObject*>o            # tbl udata
-        py_obj.runtime = <PyObject*>runtime
-        py_obj.type_flags = type_flags
-        py_obj.ref_count = 1
-        lua.luaL_getmetatable(L, POBJECT)    # tbl udata metatbl
-        lua.lua_setmetatable(L, -2)          # tbl udata
-        lua.lua_pushvalue(L, -1)             # tbl udata udata
-        pyref = _PyReference.__new__(_PyReference)
-        pyref._ref = lua.luaL_ref(L, -3)     # tbl udata
-        pyref._obj = o
-        lua.lua_remove(L, -2)                # udata
-        # originally, we just used:
-        #cpython.ref.Py_INCREF(o)
-        # now, we store an owned reference in "runtime._pyrefs_in_lua" to keep it visible to Python
-        # and a borrowed reference in "py_obj.obj" for access from Lua
-        runtime._pyrefs_in_lua[refkey] = pyref
-        return 1  # values pushed
+        if py_obj:
+            py_obj.ref_count += 1
+            lua.lua_remove(L, -2)                       # udata
+            return 1 # values pushed
+        else:
+            lua.lua_pop(L, 1)                           # tbl
+
+    py_obj = <py_object*>lua.lua_newuserdata(L, sizeof(py_object))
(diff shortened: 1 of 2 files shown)


lupa/_lupa.pyx
```diff
@@ -1035,11 +1034,8 @@ cdef class _LuaIter:
             if self._obj is None:
                 raise StopIteration
             # iterable object
-            lua.lua_rawgeti(L, lua.LUA_REGISTRYINDEX, self._obj._ref)
+            self._obj.push_lua_object(L)
             if not lua.lua_istable(L, -1):
-                if lua.lua_isnil(L, -1):
```
guidanoli, line 1040: Note that `_LuaObject.push_lua_object` already checks whether the entry in the registry is `nil`.

lupa/_lupa.pyx
```diff
@@ -1448,9 +1463,6 @@ cdef bint call_python(LuaRuntime runtime, lua_State *L, py_object* py_obj) excep
     cdef int i, nargs = lua.lua_gettop(L) - 1
     cdef tuple args
 
-    if not py_obj:
```
guidanoli, line 1451: Note that now all functions that call `call_python` already check if the object is missing.

lupa/_lupa.pyx
```diff
@@ -1413,21 +1415,34 @@ cdef tuple unpack_multiple_lua_results(LuaRuntime runtime, lua_State *L, int nar
 
 # ref-counting support for Python objects
 
+@cython.final
+@cython.internal
+@cython.freelist(8)
+cdef class _PyReference:
+    cdef object obj
+    cdef uintptr_t refcnt
```
scoder, line 1423: I wonder if the refcount could be stored in the `py_object` struct, instead of using a Python class for it?
(I guess the `WrappedObjectFlags` might get in the way, hopefully not too much…)

lupa/tests/test.py
```diff
@@ -2619,6 +2619,64 @@ def test_nil_debug_traceback(self):
         except lupa.LuaError as e:
             self.assertNotIn("stack traceback:", e.args[0])
```
scoder, line 2621:
```suggestion


```

lupa/tests/test.py
```diff
@@ -2619,6 +2619,64 @@ def test_nil_debug_traceback(self):
...
+        self.testmissingref(X(), lambda o: assign(lupa.as_attrgetter(o).a))  # __newindex (attrgetter)
+        self.testmissingref(X(), lambda o: o())                              # __call
+
+    def test_functions(self):
+        self.testmissingref({}, print)              # reflection
+        self.testmissingref({}, iter)               # iteration
+        self.testmissingref({}, enumerate)          # enumerate
+        self.testmissingref({}, lupa.as_itemgetter) # item getter protocol
+        self.testmissingref({}, lupa.as_attrgetter) # attribute getter protocol
+
```
scoder, line 2679:
```suggestion


```

lupa/_lupa.pyx
```diff
@@ -571,8 +571,7 @@ cdef class _LuaObject:
             unlock_runtime(self._runtime)
 
     @cython.final
-    cdef inline int push_lua_object(self) except -1:
-        cdef lua_State* L = self._state
+    cdef inline int push_lua_object(self, lua_State* L) except -1:
         lua.lua_rawgeti(L, lua.LUA_REGISTRYINDEX, self._ref)
```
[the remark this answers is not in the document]
guidanoli: That's not true for threads. When testing this function, I found that if we use `self._state` instead of the `L` parameter, some coroutine tests fail.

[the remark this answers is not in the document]
guidanoli: It's easy to reproduce: find calls to `lua_rawgeti` (outside `_LuaIter`) and instead of passing `L`, pass `_LuaObject._state`.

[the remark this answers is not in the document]
scoder: All right, good to know!

lupa/_lupa.pyx
guidanoli: See the new commits, how we don't need to keep the reference count anymore.

lupa/_lupa.pyx
```diff
@@ -1412,44 +1442,51 @@ cdef tuple unpack_multiple_lua_results(LuaRuntime runtime, lua_State *L, int nar
...
-            return 0  # runtime was already cleared during GC, nothing left to do
-        if len(refs) == 1:
-            del runtime._pyrefs_in_lua[obj_id]
-        else:
-            refs.pop()  # any, really
-        return 0
+        refkey = get_pyref_key(py_obj.obj, py_obj.type_flags)
+        pyref = <_PyReference>runtime._pyrefs_in_lua.pop(refkey)
+        lua.lua_getfield(L, lua.LUA_REGISTRYINDEX, PYREFST)  # tbl
+        lua.luaL_unref(L, -1, pyref._ref)                    # tbl
```
scoder, line 1463:
```suggestion
```

lupa/_lupa.pyx
```diff
@@ -1412,44 +1442,51 @@ cdef tuple unpack_multiple_lua_results(LuaRuntime runtime, lua_State *L, int nar
...
+    except (TypeError, KeyError):
+        return 0  # runtime was already cleared during GC, nothing left to do
     except:
         try: runtime.store_raised_exception(L, b'error while cleaning up a Python object')
         finally: return -1
-
+    else:
```
scoder, line 1469:
```suggestion
    else:
        lua.lua_getfield(L, lua.LUA_REGISTRYINDEX, PYREFST)  # tbl
        lua.luaL_unref(L, -1, pyref._ref)                    # tbl
```

scoder:
Thanks!

guidanoli:
> Thanks!

You're welcome!

PR #172, opened 2021-04-05
Fixed build for PyPy

guidanoli:
* tox -e py36 was resolving to CPython 3.6, not PyPy 3.6
* tox -e py now resolves to the version of Python Tox is currently
using! It's also cleaner than that echo-tr-xargs pipe.
* Source: https://tox.readthedocs.io/en/latest/example/basic.html

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -30,7 +30,7 @@ jobs:
       run: sudo apt-get install lib${{ matrix.lua-version }}-dev
 
     - name: Run tests with tox
-      run: echo "${{ matrix.python-version }}" | tr -d -c '[0-9]' | xargs -i tox -e "py{}"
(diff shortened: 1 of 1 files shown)


guidanoli:
@scoder I am not able to have PyPy 2.7 build because of deprecation issues. Could we drop the support for Python 2.7 since it's no longer supported?

.github/workflows/ci.yml
```diff
@@ -23,14 +23,14 @@ jobs:
         python-version: ${{ matrix.python-version }}
 
     - name: Set up Python packages
-      run: python -m pip install -U tox virtualenv
+      run: python -m pip install -U pip tox virtualenv
```
scoder, line 26:
```suggestion
      run: python -m pip install -U pip<21 tox virtualenv
```

guidanoli (commented):
Seems fine to me!

PR #175, opened 2021-04-10
Fix crash on table lookups

### LINKED ISSUE #174: Error thrown in 'index' metamethod not caught
**What is it?**
`_LuaObject.__getitem__` is unsafe because it calls `lua_gettable`, which may call a `__index` tag method, which in turn, might throw an error (a long jump in C) in the middle of Python code.

**How to reproduce it?**
```python
from lupa import LuaRuntime
lua = LuaRuntime()
lua.execute('''
t = {}
setmetatable(t, {__index = function() error('indexing error!') end})
''')
lua_t = lua.eval('t')
print(lua_t[1])
# Segmentation fault
```

**How to fix it?**
We could call `lua_rawget` instead of `lua_gettable`, which would not call the `index` metamethod.
Or, if keeping Lua semantics in Python is very important, we could call the `index` metamethod in protected mode.

guidanoli:
* Adds `_LuaObject._gettable`, which is a safer version `lua_gettable` which doesn't make long jumps
* Modifies `_LuaObject._getitem` to call `_LuaObject._gettable` instead of `lua_gettable`
* Adds tests for errors thrown by the "index" metamethod

Closes https://github.com/scoder/lupa/issues/174

### DIFF
--- a/lupa/_lupa.pyx
+++ b/lupa/_lupa.pyx
@@ -678,47 +678,6 @@ cdef class _LuaObject:
     def __getitem__(self, index_or_name):
         return self._getitem(index_or_name, is_attr_access=False)
 
-    @cython.final
-    cdef int _get_from_table(self, int table, int key):
-        cdef lua_State* L = self._state
-        cdef int top
-        lua.lua_pushvalue(L, key)                     # key
-        if lua.lua_type(L, table) == lua.LUA_TTABLE:
-            lua.lua_rawget(L, table)                  # value
-            if not lua.lua_isnil(L, -1):
-                return 1
-            lua.lua_pop(L, 1)                         #
-            if not lua.lua_getmetatable(L, table):    # mt?
-                lua.lua_pushnil(L)                    # nil
-                return 1
-            lua.lua_pushlstring(L, "__index", 7)      # mt tm
-            top = lua.lua_gettop(L)
-            if not self._get_from_table(top - 1, top):      # mt[tm]
-                return 0                              #
-            if lua.lua_isnil(L, -1):
-                return 1                              # nil
-        else:
-            if not lua.lua_getmetatable(L, table):    # mt?
-                lua.lua_pushnil(L)                    # nil
-                return 1
-            lua.lua_pushlstring(L, "__index", 7)      # mt tm
-            top = lua.lua_gettop(L)
-            if not self._get_from_table(top - 1, top):      # mt[tm]
-                return 0                              #
-            if lua.lua_isnil(L, -1):
-                return 0                              # nil
-
-        if lua.lua_type(L, -1) == lua.LUA_TFUNCTION:
-            lua.lua_pushvalue(L, table)               # h tbl
-            lua.lua_pushvalue(L, key)                 # h tbl key
-            if lua.lua_pcall(L, 2, 1, 0):             # h(tbl, key)
-                lua.lua_pop(L, 1)                     # err
-                return 0                              #
-            return 1                                  # val
-        else:
-            lua.lua_pushvalue(L, key)                 # h key
-            top = lua.lua_gettop(L)
-            return self._get_from_table(top - 1, top)       # h[key]
 
     @cython.final
     cdef _getitem(self, name, bint is_attr_access):
@@ -734,8 +693,7 @@ cdef class _LuaObject:
                     "item/attribute access not supported on functions")
             # table[nil] fails, so map None -> python.none for Lua tables
             py_to_lua(self._runtime, L, name, wrap_none=lua_type == lua.LUA_TTABLE)
-            if not self._get_from_table(old_top + 1, old_top + 2):
-                raise LuaError("table field access error")
+            luaP_gettable(self._runtime, L, -2)
             return py_from_lua(self._runtime, L, -1)
         finally:
             lua.lua_settop(L, old_top)
@@ -1959,3 +1917,35 @@ cdef void luaL_openlib(lua_State *L, const char *libname,
         luaL_setfuncs(L, l, nup)
     else:
         lua.lua_pop(L, nup)
+
+
+# protected calls for manipulating the Lua stack in Python
+
+cdef int luaP_gettable_aux(lua_State* L) nogil:
+    lua.lua_pushvalue(L, lua.lua_upvalueindex(1))  # key
+    lua.lua_gettable(L, lua.lua_upvalueindex(2))   # tbl[key]
+    return 1
+
+
+cdef luaP_gettable(LuaRuntime runtime, lua_State* L, int t):
+    """Protected call to ``lua_gettable``
+
+    Pushes onto the stack the value ``tbl[key]``, where
+    ``tbl`` is the value at the given valid index ``t`` and
+    ``key`` is the value at the top of the stack.
+
+    On failure, an error is raised.
+    """
+    cdef int result
+    if not lua.lua_checkstack(L, 1):
+        raise MemoryError()
+                                                   # key
+    lua.lua_pushvalue(L, t)                        # key tbl
+    lua.lua_pushcclosure(L, luaP_gettable_aux, 2)  # aux
+    result = lua.lua_pcall(L, 0, 1, 0)
+    if result:
+        try:                                       # err
+            raise_lua_error(runtime, L, result)
+        finally:
+            lua.lua_pop(L, 1)                      #
+                                                   # tbl[key]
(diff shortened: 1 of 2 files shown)


lupa/_lupa.pyx
```diff
@@ -685,7 +726,8 @@ cdef class _LuaObject:
                     "item/attribute access not supported on functions")
             # table[nil] fails, so map None -> python.none for Lua tables
             py_to_lua(self._runtime, L, name, wrap_none=lua_type == lua.LUA_TTABLE)
-            lua.lua_gettable(L, -2)
+            if not self._gettable(old_top + 1, old_top + 2):
+                raise LuaError("table field access error")
```
scoder, line 730: Can't we wrap `lua_gettable()` in a C function that uses `setjmp()` to catch the error?

guidanoli: It's not clear to me how would we be able to catch the error bzw the call to `longjmp` in C.

scoder: Ok, I read some more on error handling in Lua C-API code. Seems that Lua's `longjmp` error handling is not exposed by itself and things _are_ really complicated. `lua_atpanic()` seems an option here, but requires some form of runtime global C state to figure out where to jump back to. Or use protected mode, as you did. :-[

scoder:
IIUC, you are trying to find out if there is an `__index` method that may cause problems, and if so, call it in protected mode. Otherwise, you try to find cases that are safe and handle those internally. Is that correct?

More readable comments than the cryptic "mt", "tm", "h" abbreviations would help to distinguish the different cases.

lupa/_lupa.pyx
guidanoli: Right, currently we call `lua_atpanic(1)`, altough I am not quite sure what it does when passing 1 (is it stack indexation?).
I believe the only safe way is to call in protected mode, unfortunately. I've read in http://lua-users.org/wiki/ErrorHandlingBetweenLuaAndCplusplus that we could compile Lua so that instead of C `longjmp`, C++ exceptions are raised. But I guess we could only do that with bundled Lua, not with pkg-config Lua or custom Lua.

guidanoli:
> IIUC, you are trying to find out if there is an `__index` method that may cause problems, and if so, call it in protected mode. Otherwise, you try to find cases that are safe and handle those internally. Is that correct?
> 
> More readable comments than the cryptic "mt", "tm", "h" abbreviations would help to distinguish the different cases.

Yes, I basically followed the pseudo-code for the "index" metamethod in the [Lua reference manual](https://www.lua.org/manual/5.1/manual.html#2.8) and mapped it to the Lua-C API.

Moreover, I'll try to find better abbreviations for the comments.

lupa/_lupa.pyx
```diff
@@ -671,6 +671,47 @@ cdef class _LuaObject:
     def __getitem__(self, index_or_name):
         return self._getitem(index_or_name, is_attr_access=False)
 
+    @cython.final
+    cdef int _gettable(self, int table, int key):
```
scoder, line 675:
```suggestion
    cdef int _get_from_table(self, int table, int key):
```

lupa/_lupa.pyx
```diff
@@ -671,6 +671,47 @@ cdef class _LuaObject:
...
+            lua.lua_pushvalue(L, table)               # h tbl
+            lua.lua_pushvalue(L, key)                 # h tbl key
+            if lua.lua_pcall(L, 2, 1, 0):             # h(tbl, key)
+                lua.lua_pop(L, 1)                     # err
+                return 0                              #
+            return 1                                  # val
+        else:
+            lua.lua_pushvalue(L, key)                 # h key
+            top = lua.lua_gettop(L)
+            return self._gettable(top - 1, top)       # h[key]
```
scoder, line 713:
```suggestion
            return self._get_from_table(top - 1, top)       # h[key]
```

lupa/_lupa.pyx
```diff
@@ -671,6 +671,47 @@ cdef class _LuaObject:
...
+            lua.lua_rawget(L, table)                  # value
+            if not lua.lua_isnil(L, -1):
+                return 1
+            lua.lua_pop(L, 1)                         #
+            if not lua.lua_getmetatable(L, table):    # mt?
+                lua.lua_pushnil(L)                    # nil
+                return 1
+            lua.lua_pushlstring(L, "__index", 7)      # mt tm
+            top = lua.lua_gettop(L)
+            if not self._gettable(top - 1, top):      # mt[tm]
```
scoder, line 689:
```suggestion
            if not self._get_from_table(top - 1, top):      # mt[tm]
```

lupa/_lupa.pyx
```diff
@@ -671,6 +671,47 @@ cdef class _LuaObject:
...
+                return 0                              #
+            if lua.lua_isnil(L, -1):
+                return 1                              # nil
+        else:
+            if not lua.lua_getmetatable(L, table):    # mt?
+                lua.lua_pushnil(L)                    # nil
+                return 1
+            lua.lua_pushlstring(L, "__index", 7)      # mt tm
+            top = lua.lua_gettop(L)
+            if not self._gettable(top - 1, top):      # mt[tm]
```
scoder, line 699:
```suggestion
            if not self._get_from_table(top - 1, top):      # mt[tm]
```

lupa/_lupa.pyx
```diff
@@ -685,7 +726,8 @@ cdef class _LuaObject:
                     "item/attribute access not supported on functions")
             # table[nil] fails, so map None -> python.none for Lua tables
             py_to_lua(self._runtime, L, name, wrap_none=lua_type == lua.LUA_TTABLE)
-            lua.lua_gettable(L, -2)
+            if not self._gettable(old_top + 1, old_top + 2):
```
scoder, line 729:
```suggestion
            if not self._get_from_table(old_top + 1, old_top + 2):
```

scoder (commented):
The stack juggling is awful, but I can't see how to do better. Let's help the innocent reader with some comments and get it merged.

guidanoli:
Doing some adjustments still

guidanoli:
@scoder should be good to go

lupa/_lupa.pyx
```diff
@@ -1916,3 +1917,35 @@ cdef void luaL_openlib(lua_State *L, const char *libname,
...
+
+# protected calls for manipulating the Lua stack in Python
+
+cdef int luaP_gettable_aux(lua_State* L) nogil:
+    lua.lua_pushvalue(L, lua.lua_upvalueindex(1))  # key
+    lua.lua_gettable(L, lua.lua_upvalueindex(2))   # tbl[key]
+    return 1
+
+
+cdef luaP_gettable(LuaRuntime runtime, lua_State* L, int t):
```
scoder: Please don't use `lua…` as prefix for module internal functions. That's confusing because they look like Lua API functions. `_get_from_lua_table` seems an appropriate name. Also, please give `t` a name that suggests a reference to a stack index, e.g. `table_ref` or `table_on_stack` or something better that I'm sure you can come up with.

Actually, why is `t` needed at all? Can't we just let the caller make sure that the stack contains [table, key]` in the right order? After all, this is only used in a single place (for now).

guidanoli: Agreed!

lupa/tests/test.py
```diff
@@ -2620,6 +2620,19 @@ def test_nil_debug_traceback(self):
...
+# tests for table access error
+
+class TestTableAccessError(SetupLuaRuntimeMixin, unittest.TestCase):
+    def test_error_index_metamethod(self):
+        self.lua.execute('''
+        t = {}
+        setmetatable(t, {__index = function() error() end})
+        ''')
+        lua_t = self.lua.eval('t')
+        self.assertRaises(lupa.LuaError, lambda t, k: t[k], lua_t, 'k')
```
scoder, line 2633: Can we somehow make sure that the `__index` method was really called, and that the error we receive is the one that we intentionally raised and nothing else? Otherwise, it's too easy to break something somewhere and still receive some kind of `LuaError` here, but a different one.

scoder (commented):
The implementation looks good to me now.

lupa/tests/test.py
guidanoli: Sure!

guidanoli: Done in 58d2883

scoder:
Thanks!

PR #177, opened 2021-04-11
Add python.args function for building keyword argument dicts

### LINKED ISSUE #176: Alternative way to pass keyword arguments to Python functions in Lua
**Issue:**
There is no clean way to call an arbitrary Python method/function with keyword arguments in Lua.

**Current situation:**
Currently the only way to do so would be by evaluating a Python lambda function in Lua which would somehow convert some positional arguments into keyword arguments, but that's not only ugly but inefficient.

**Suggestion:**
If we want to call a function with keyword arguments, there should be no need to wrap the function in some type of decorator (like is done currently with `unpacks_lua_table` and `unpacks_lua_table_method`). This should be done in the function call, for example, with a table constructor with a unique metatable which identifies a table as a set of keyword arguments:

```lua
python.builtins.print(1, 2, 3, python.kwargs{sep=', '}) -- Prints "1, 2, 3"
python.builtins.dict(python.kwargs{a=1, b=2})           -- {"a": 1, "b": 2}
```

guidanoli:
* Add `python.args` function for building Python arguments (+tests)
* Add internal `unpack_lua_table` used by `lupa.unpacks_lua_table`, `lupa.unpacks_lua_table_method` and `python.args`
  * Throw `IndexError` if table being unpacked has "isolated" integer key (+tests)

Closes https://github.com/scoder/lupa/issues/176

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -10,6 +10,8 @@ Upcoming release
 * The Lua implementation name and version string is provided as
   ``LuaRuntime.lua_implementation``.
 
+* GH#177: Tables that are not sequences raise ``IndexError`` when unpacked.
+
 
 1.9 (2019-12-21)
 ----------------
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -10,7 +10,11 @@ Upcoming release
 * The Lua implementation name and version string is provided as
   ``LuaRuntime.lua_implementation``.
 
+* GH#177: A new function ``python.args(*args, **kwargs)`` was added to help with building
+  Python argument tuples and keyword argument dicts for Python function calls from Lua code.
+
 * GH#177: Tables that are not sequences raise ``IndexError`` when unpacked.
+  They previously raised a ``TypeError``.
 
 
 1.9 (2019-12-21)
--- a/lupa/_lupa.pyx
+++ b/lupa/_lupa.pyx
@@ -64,7 +64,6 @@ except ImportError:
 DEF POBJECT = b"POBJECT" # as used by LunaticPython
 DEF PYREFST = b"LUPA_PYTHON_REFERENCES_TABLE"
 DEF LUART = b"LUPA_LUA_RUNTIME"
-DEF PYARGS = b"LUPA_PYTHON_ARGUMENTS_METATABLE"
 
 cdef extern from *:
     """
@@ -449,12 +448,8 @@ cdef class LuaRuntime:
         lua.lua_pop(L, 1)
 
         # register the Lua runtime
-        lua.lua_pushlightuserdata(L, <void*><PyObject*>self)
+        lua.lua_pushlightuserdata(L, <void*>self)
         lua.lua_setfield(L, lua.LUA_REGISTRYINDEX, LUART)
-
-        # register the Python arguments metatable
-        lua.lua_newtable(L)
-        lua.lua_setfield(L, lua.LUA_REGISTRYINDEX, PYARGS)
 
         # create and store the python references table
         lua.lua_newtable(L)                                  # tbl
@@ -581,12 +576,14 @@ cdef class _LuaObject:
             raise LuaError("lost reference")
 
     def __eq__(self, o):
+        assert self._runtime is not None
         if not isinstance(o, _LuaObject):
             return False
-        cdef _LuaObject other = <_LuaObject>o
-        if self._state != other._state:
-            return False
         cdef lua_State* L = self._state
+        cdef _LuaObject other = <_LuaObject>o
+        if L != other._state:
+            return False
+        lock_runtime(self._runtime)
         old_top = lua.lua_gettop(L)
         cdef bint equal = False
         try:
@@ -595,6 +592,7 @@ cdef class _LuaObject:
             equal = lua.lua_rawequal(L, -1, -2)
         finally:
             lua.lua_settop(L, old_top)
+            unlock_runtime(self._runtime)
         return equal
 
     def __call__(self, *args):
@@ -837,19 +835,22 @@ cdef class _LuaTable(_LuaObject):
         and a dictionary of keyword arguments
         """
         assert self._runtime is not None
-        cdef long length = <long>self._len()
-        cdef bytes encoding = self._runtime._source_encoding
-        cdef tuple args = cpython.tuple.PyTuple_New(<Py_ssize_t>length)
+        cdef tuple args
         cdef dict kwargs = {}
+        cdef bytes encoding
+        cdef Py_ssize_t index, length
         lock_runtime(self._runtime)
         try:
+            encoding = self._runtime._source_encoding
+            length = <Py_ssize_t>self._len()
+            args = cpython.tuple.PyTuple_New(length)
             for key, value in self.items():
                 if isinstance(key, (int, long)) and not isinstance(key, bool):
-                    if key >= 1 and key <= length:
-                        cpython.ref.Py_INCREF(value)
-                        cpython.tuple.PyTuple_SET_ITEM(args, <Py_ssize_t>(key-1), value)
-                    else:
+                    index = <Py_ssize_t>key
+                    if index < 1 or index > length:
                         raise IndexError("table index out of range")
+                    cpython.ref.Py_INCREF(value)
+                    cpython.tuple.PyTuple_SET_ITEM(args, index-1, value)
                 elif isinstance(key, bytes):
                     if IS_PY2:
                         kwargs[key] = value
@@ -1875,7 +1876,7 @@ cdef int py_iter_next_with_gil(lua_State* L, py_object* py_iter) with gil:
         try: runtime.store_raised_exception(L, b'error while calling next(iterator)')
         finally: return -1
 
-# support for calling Python objects in Lua with keyword arguments
+# support for calling Python objects in Lua with Python-like arguments
 
 cdef class _PyArguments:
     cdef tuple args
@@ -1895,12 +1896,11 @@ cdef int py_args_with_gil(lua_State* L) with gil:
         finally: return -1
 
 cdef int py_args(lua_State* L) nogil:
-    lua.luaL_checktype(L, 1, lua.LUA_TTABLE)  # tbl ...
-    lua.lua_settop(L, 1)                      # tbl
+    lua.luaL_checktype(L, 1, lua.LUA_TTABLE)
     result = py_args_with_gil(L)
     if result < 0:
         return lua.lua_error(L) # never returns!
-    return result                             # args
+    return result
 
 # 'python' module functions in Lua
 
--- a/lupa/_lupa.pyx
+++ b/lupa/_lupa.pyx
@@ -38,8 +38,8 @@ cdef extern from *:
     """
     ctypedef size_t uintptr_t
 
-cdef object exc_info
-from sys import exc_info
+cdef object exc_info, exit
+from sys import exc_info, exit
 
 cdef object Mapping
 try:
@@ -63,7 +63,6 @@ except ImportError:
 
 DEF POBJECT = b"POBJECT" # as used by LunaticPython
 DEF PYREFST = b"LUPA_PYTHON_REFERENCES_TABLE"
-DEF LUART = b"LUPA_LUA_RUNTIME"
 
 cdef extern from *:
     """
@@ -247,8 +246,7 @@ cdef class LuaRuntime:
 
         lua.luaL_openlibs(L)
         self.init_python_lib(register_eval, register_builtins)
-        lua.lua_settop(L, 0)
-        lua.lua_atpanic(L, <lua.lua_CFunction>1)
+        lua.lua_atpanic(L, panic)
 
     def __dealloc__(self):
         if self._state is not NULL:
@@ -441,23 +439,24 @@ cdef class LuaRuntime:
     cdef int init_python_lib(self, bint register_eval, bint register_builtins) except -1:
         cdef lua_State *L = self._state
 
-        # create 'python' lib and register our own object metatable
-        luaL_openlib(L, "python", py_lib, 0)
-        lua.luaL_newmetatable(L, POBJECT)
-        luaL_openlib(L, NULL, py_object_lib, 0)
-        lua.lua_pop(L, 1)
+        # create 'python' lib
+        luaL_openlib(L, "python", py_lib, 0)       # lib
+        lua.lua_pushlightuserdata(L, <void*>self)  # lib udata
+        lua.lua_pushcclosure(L, py_args, 1)        # lib function
+        lua.lua_setfield(L, -2, "args")            # lib 
 
-        # register the Lua runtime
-        lua.lua_pushlightuserdata(L, <void*>self)
-        lua.lua_setfield(L, lua.LUA_REGISTRYINDEX, LUART)
+        # register our own object metatable
+        lua.luaL_newmetatable(L, POBJECT)          # lib metatbl
+        luaL_openlib(L, NULL, py_object_lib, 0)
+        lua.lua_pop(L, 1)                          # lib 
 
         # create and store the python references table
-        lua.lua_newtable(L)                                  # tbl
-        lua.lua_createtable(L, 0, 1)                         # tbl metatbl
-        lua.lua_pushlstring(L, "v", 1)                       # tbl metatbl "v"
-        lua.lua_setfield(L, -2, "__mode")                    # tbl metatbl
-        lua.lua_setmetatable(L, -2)                          # tbl
-        lua.lua_setfield(L, lua.LUA_REGISTRYINDEX, PYREFST)  #
+        lua.lua_newtable(L)                                  # lib tbl
+        lua.lua_createtable(L, 0, 1)                         # lib tbl metatbl
+        lua.lua_pushlstring(L, "v", 1)                       # lib tbl metatbl "v"
+        lua.lua_setfield(L, -2, "__mode")                    # lib tbl metatbl
+        lua.lua_setmetatable(L, -2)                          # lib tbl
+        lua.lua_setfield(L, lua.LUA_REGISTRYINDEX, PYREFST)  # lib 
 
         # register global names in the module
         self.register_py_object(b'Py_None',  b'none', None)
@@ -465,6 +464,9 @@ cdef class LuaRuntime:
             self.register_py_object(b'eval',     b'eval', eval)
         if register_builtins:
             self.register_py_object(b'builtins', b'builtins', builtins)
+
+        # pop 'python' lib
+        lua.lua_pop(L, 1)
 
         return 0  # nothing left to return on the stack
 
@@ -1882,11 +1884,9 @@ cdef class _PyArguments:
     cdef tuple args
     cdef dict kwargs
 
-cdef int py_args_with_gil(lua_State* L) with gil:
-    cdef LuaRuntime runtime
+cdef int py_args_with_gil(LuaRuntime runtime, lua_State* L) with gil:
     cdef _PyArguments pyargs
     try:
-        runtime = luaL_getruntime(L)
         table = new_lua_table(runtime, L, 1)
         pyargs = _PyArguments.__new__(_PyArguments)
         pyargs.args, pyargs.kwargs = table._unpack()
@@ -1896,8 +1896,13 @@ cdef int py_args_with_gil(lua_State* L) with gil:
         finally: return -1
 
 cdef int py_args(lua_State* L) nogil:
+    cdef PyObject* runtime
+    runtime = <PyObject*>lua.lua_touserdata(L, lua.lua_upvalueindex(1))
+    if not runtime:
+        return lua.luaL_error(L, "missing runtime")
     lua.luaL_checktype(L, 1, lua.LUA_TTABLE)
-    result = py_args_with_gil(L)
+    with gil:
+        result = py_args_with_gil(<LuaRuntime>runtime, L)
     if result < 0:
         return lua.lua_error(L) # never returns!
     return result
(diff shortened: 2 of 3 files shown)


lupa/_lupa.pyx
```diff
@@ -1799,6 +1823,15 @@ cdef int py_iter_next_with_gil(lua_State* L, py_object* py_iter) with gil:
         try: runtime.store_raised_exception(L, b'error while calling next(iterator)')
         finally: return -1
 
+# keyword argument support for Python objects in Lua
+
+cdef int py_kwargs(lua_State* L) nogil:
```
scoder, line 1828: Why not construct the Python dict right here? We should do ahead of time whatever we can. If anything, then the call is the critical part, not the keyword building. Only if more than one kwargs dict is found in a call, then we need to do something (i.e. create a merged superset).

The object that we create here can be entirely opaque. Only Lupa needs to understand its intention (i.e. passing it as kwargs when found in a call).

lupa/_lupa.pyx
```diff
@@ -1446,20 +1451,48 @@ cdef int py_object_gc(lua_State* L) nogil:
...
 
-    if not nargs:
+    if nargs == 0:
         lua.lua_settop(L, 0)  # FIXME
         result = f()
     else:
-        arg = py_from_lua(runtime, L, 2)
+        kwargs = {}
+        
+        if lua.lua_istable(L, -1) and lua.lua_getmetatable(L, -1):
```
scoder, line 1470: This part looks mostly like `_fix_args_kwargs()` spelled differently. I don't see a need for the duplication.

scoder (commented):
I'm not convinced that this is needed. Let's discuss it in #176.

lupa/_lupa.pyx
guidanoli: I disagree with having to merge kwargs dict. That just doesn't seem natural. This would also create another problem: If more than one kwargs dict is found in the call, which dict has more priority? Adding ambiguity is undesired.

```lua
python.print(1, 2, 3, python.kwargs{sep=', '}, python.kwargs{end='!\n'}, python.kwargs{sep='\t'})
```

lupa/_lupa.pyx
guidanoli: This is way different than `_fix_args_kwargs`.

`_fix_args_kwargs` checks if the first argument is a table and unpacks it.
`call_python` checks if the last argument is a table with our hidden metatable and unpacks it.

lupa/_lupa.pyx
scoder: > I disagree with having to merge kwargs dict

The use case becomes clearer as soon as you diverge from using literal keyword arguments. Imagine a (contrived) call like
```lua
python.builtins.print(x, TO_FILE, AS_TSV)
```
where `TO_FILE = kwargs{file=some_file)` and `AS_TSV = kwargs{sep=";"}`.

> which dict has more priority

```python
>>> print(1,2,3, **dict(sep='--'), **dict(sep=':'))
  File "<stdin>", line 1, in <module>
TypeError: print() got multiple values for keyword argument 'sep'
```

lupa/_lupa.pyx
scoder: `_fix_args_kwargs` takes a Lua table and splits it into positional and keyword arguments for a Python call. `call_python` (after your change) takes a Lua table and splits it into positional and keyword arguments for a Python call. Where exactly is the difference?

lupa/_lupa.pyx
guidanoli: I do understand your point now, but I need to research on how to replicate the keyword arguments expansion of multiple dictionaries or the merging of many mutually-exclusive dictionaries using the Python-C API.

lupa/_lupa.pyx
guidanoli: On `_fix_args_kwargs`, table entries accept integer keys as positional arguments, while on `call_python` (after the change), table entries must have string entries only.

guidanoli: But we could create a `_LuaTable.to_dict` method which converts it to a dictionary, so that we could soak the repeated code from both functions, what do you think?

lupa/_lupa.pyx
scoder: Luckily, Lupa is written in Cython, not C. :)

guidanoli: You're right. We could do something in the lines of...

```cython
new_dict = dict(**dict_a, **dict_b)
```

...since `dict` is a callable so it would raise a `TypeError`.

lupa/_lupa.pyx
scoder: sorry, right, in the case of `call_python`, the positional arguments come from the stack, not from a table. And as I said, it's better to prepare the call dict upfront.

Why not have a general function `python.pyargs(lua_table)` that returns an object that a Python call unpacks into an args/kwargs pair? Then you could just say
```lua
python.builtins.print(python.pyargs{1, 2, 3, sep=";"})
```
To me, that seems more versatile and more readable than requiring to split the arguments into "normal" Lua arguments and "special" Python keyword arguments _within the same call_.

guidanoli: The only problem I find with this is that merging positional arguments seems inappropriate, unless we are willing to admit this is the only argument being passed.

scoder:
```python
>>> print(*(1,2,3), *(4,5,6))
1 2 3 4 5 6
```

guidanoli: Fair enough.

lupa/_lupa.pyx
```diff
@@ -819,6 +829,41 @@ cdef class _LuaTable(_LuaObject):
...
+        and a dictionary of keyword arguments
+        """
+        assert self._runtime is not None
+        cdef tuple args
+        cdef dict kwargs = {}
+        cdef bytes encoding
+        cdef Py_ssize_t index, length
+        lock_runtime(self._runtime)
+        try:
+            encoding = self._runtime._source_encoding
```
scoder, line 844: Wow, this is literally the only thing that we need the `LuaRuntime` for here. Isn't there another way to get the encoding than to put the `LuaRuntime` in the Lua registry?

lupa/_lupa.pyx
```diff
@@ -819,6 +829,41 @@ cdef class _LuaTable(_LuaObject):
...
+                    index = <Py_ssize_t>key
+                    if index < 1 or index > length:
+                        raise IndexError("table index out of range")
+                    cpython.ref.Py_INCREF(value)
+                    cpython.tuple.PyTuple_SET_ITEM(args, index-1, value)
+                elif isinstance(key, bytes):
+                    if IS_PY2:
+                        kwargs[key] = value
+                    else:
+                        kwargs[(<bytes>key).decode(encoding)] = value
```
scoder, line 858: BTW, I think this is (and was) buggy. `key` can only be `bytes` here if the source encoding is None (otherwise it already got decoded in `py_from_lua()`), and in that case, `.decode()` will reject the invalid argument. We can probably just ignore the `bytes` case. But `py_from_lua()` still needs the encoding.

guidanoli: But `self._source_encoding is not None` because (on `LuaRuntime.__cinit__`)
```cython
self._source_encoding = _asciiOrNone(source_encoding) or self._encoding or b'UTF-8'
```

lupa/_lupa.pyx
guidanoli: We need the `LuaRuntime` for building the `_LuaTable` and converting its keys and values. If we opt to write a `LuaRuntime`-less `_unpacks`, we'd have to duplicate a lot of code.

lupa/_lupa.pyx
```diff
@@ -1896,8 +1896,13 @@ cdef int py_args_with_gil(lua_State* L) with gil:
         finally: return -1
 
 cdef int py_args(lua_State* L) nogil:
+    cdef PyObject* runtime
+    runtime = <PyObject*>lua.lua_touserdata(L, lua.lua_upvalueindex(1))
```
scoder, line 1900: That's much better, good idea.

lupa/_lupa.pyx
```diff
@@ -1896,8 +1896,13 @@ cdef int py_args_with_gil(lua_State* L) with gil:
...
+    if not runtime:
+        return lua.luaL_error(L, "missing runtime")
     lua.luaL_checktype(L, 1, lua.LUA_TTABLE)
-    result = py_args_with_gil(L)
+    with gil:
+        result = py_args_with_gil(<LuaRuntime>runtime, L)
```
scoder, line 1905: Just pass the `PyObject*` and cast it to `<LuaRuntime>` inside of the `with gil` function. That avoids the extra (and in the face of longjmp, risky) `with gil` block here. You can also use `<LuaRuntime?>` as a type cast to get an additional Python type check.

lupa/_lupa.pyx
```diff
@@ -1916,3 +1993,9 @@ cdef void luaL_openlib(lua_State *L, const char *libname,
         luaL_setfuncs(L, l, nup)
     else:
         lua.lua_pop(L, nup)
+
+
+cdef int panic(lua_State* L) with gil:
+    exit("Lua panicked and exited")
```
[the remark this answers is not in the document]
guidanoli: You're right. I'll think of a better panic function.

lupa/_lupa.pyx
```diff
@@ -1487,17 +1534,27 @@ cdef int py_object_gc(lua_State* L) nogil:
...
 
     f = <object>py_obj.obj
 
-    if not nargs:
+    if nargs == 0:
         lua.lua_settop(L, 0)  # FIXME
         result = f()
     else:
-        arg = py_from_lua(runtime, L, 2)
+        args = []
```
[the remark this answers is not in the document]
guidanoli: Great advise!

lupa/_lupa.pyx
```diff
@@ -585,6 +577,26 @@ cdef class _LuaObject:
...
+        cdef bint equal = False
+        try:
+            self.push_lua_object(L)
+            other.push_lua_object(L)
+            equal = lua.lua_rawequal(L, -1, -2)
+        finally:
+            lua.lua_settop(L, old_top)
+            unlock_runtime(self._runtime)
+        return equal
+
```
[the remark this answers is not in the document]
guidanoli: Yes, but this feature comes in extremely handy for `PythonArgumentsInLuaTest` class in `lupa/tests/test.py`, when testing if positional and keyword arguments are correctly passed from Lua to Python :)

lupa/_lupa.pyx
guidanoli: Thanks!

lupa/_lupa.pyx
guidanoli: Thanks!

lupa/_lupa.pyx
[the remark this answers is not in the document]
guidanoli: Solved in e4289f2

lupa/_lupa.pyx
```diff
@@ -1556,16 +1640,26 @@ cdef bint call_python(LuaRuntime runtime, lua_State *L, py_object* py_obj) excep
...
+        args = ()
+        kwargs = {}
+        
+        for i in range(nargs):
+            arg = py_from_lua(runtime, L, i+2)
+            if isinstance(arg, _PyArguments):
+                args += (<_PyArguments>arg).args
+                kwargs = dict(**kwargs, **(<_PyArguments>arg).kwargs)
+            else:
+                args += (arg, )
```
scoder, line 1660: Hmm. Now that I see the `args += (arg,)` – most of the time, there won't be any `_PyArguments` on the stack. Building an args tuple here one by one seems rather wasteful in that regard.

I think you were right with your original approach to build a list instead, and just use `.append()` and `.extend()`.

Alternatively, there is a documented [`_PyTuple_Resize()`](https://docs.python.org/3/c-api/tuple.html#c._PyTuple_Resize) function which would fit this use case, while allowing us to allocate a pre-sized tuple. It's also supported by PyPy, I think.

lupa/_lupa.pyx
```diff
@@ -1898,6 +1983,35 @@ cdef int py_iter_next_with_gil(lua_State* L, py_object* py_iter) with gil:
...
+cdef int py_args_with_gil(PyObject* runtime_obj, lua_State* L) with gil:
+    cdef _PyArguments pyargs
+    cdef LuaRuntime runtime
+    try:
+        runtime = <LuaRuntime?>runtime_obj
+        pyargs = _PyArguments.__new__(_PyArguments)
+        pyargs.args, pyargs.kwargs = unpack_lua_table(runtime)
+        return py_to_lua_custom(runtime, L, pyargs, 0)
+    except:
+        try: runtime.store_raised_exception(L, b'error while calling args')
```
scoder, line 2001:
```suggestion
        try: runtime.store_raised_exception(L, b'error while calling python.args()')
```

lupa/_lupa.pyx
```diff
@@ -1997,3 +2111,5 @@ cdef void luaL_openlib(lua_State *L, const char *libname,
         luaL_setfuncs(L, l, nup)
     else:
         lua.lua_pop(L, nup)
+
+
```
scoder, line 2115:
```suggestion
```

lupa/tests/test.py
```diff
@@ -2272,8 +2274,8 @@ def test_posargs_kwargs(self):
 
     def test_posargs_kwargs_bad(self):
         self.assertIncorrect(self.arg2, "{5, y=6, z=7}")
-        self.assertIncorrect(self.arg2, "{5, [3]=6}")
-        self.assertIncorrect(self.arg2, "{x=5, [2]=6}")  # I guess it's ok to reject this
+        self.assertIncorrect(self.arg2, "{5, [3]=6}", error=IndexError)
+        self.assertIncorrect(self.arg2, "{x=5, [2]=6}", error=IndexError)
```
scoder, line 2278: Hmm, seems worth mentioning this in the changelog. It's an incompatible change.

lupa/tests/test.py
```diff
@@ -2620,6 +2622,115 @@ def test_nil_debug_traceback(self):
...
+            if objtype not in {'number', 'string'}:
+                self.assertIncorrect('python.args{[kwargs["%s"]] = true}' % objtype,
+                        regex='table key is neither an integer nor a string')
+ 
+    def test_kwargs_merge(self):
+        self.assertResult('python.args{1, a=1}, python.args{2}, python.args{}, python.args{b=2}', (1, 2), dict(a=1, b=2))
+
+    def test_kwargs_merge_conflict(self):
+        self.assertIncorrect('python.args{a=1}, python.args{a=2}', regex='multiple values')
+
```
scoder, line 2714:
```suggestion


```

lupa/tests/test.py
```diff
@@ -2800,7 +2911,7 @@ def test_functions(self):
         self.testmissingref({}, enumerate)          # enumerate
         self.testmissingref({}, lupa.as_itemgetter) # item getter protocol
         self.testmissingref({}, lupa.as_attrgetter) # attribute getter protocol
-
+
```
scoder, line 2917:
```suggestion

```

lupa/_lupa.pyx
```diff
@@ -535,6 +547,89 @@ def unpacks_lua_table_method(meth):
     return wrapper
 
 
+cdef int check_lua_stack(lua_State* L, int extra) except -1:
+    """Wrapper around lua_checkstack
+    On failure, a LuaMemoryError is raised
```
scoder, line 552:
```suggestion
    """Wrapper around lua_checkstack.
    On failure, a LuaMemoryError is raised.
```

lupa/tests/test.py
```diff
@@ -2620,6 +2622,115 @@ def test_nil_debug_traceback(self):
...
+################################################################################
+# tests for keyword arguments
+
+class PythonArgumentsInLuaTest(SetupLuaRuntimeMixin, unittest.TestCase):
+
+    def __init__(self, *args, **kwargs):
+        super(PythonArgumentsInLuaTest, self).__init__(*args, **kwargs)
+        self.get_args = lambda *args, **kwargs: args
+        self.get_kwargs = lambda *args, **kwargs: kwargs
+        self.get_none = lambda *args, **kwargs: None
```
scoder, line 2634: Would a `staticmethod` also work here?

lupa/tests/test.py
```diff
@@ -2620,6 +2622,115 @@ def test_nil_debug_traceback(self):
...
+    def test_kwargs_merge(self):
+        self.assertResult('python.args{1, a=1}, python.args{2}, python.args{}, python.args{b=2}', (1, 2), dict(a=1, b=2))
+
+    def test_kwargs_merge_conflict(self):
+        self.assertIncorrect('python.args{a=1}, python.args{a=2}', regex='multiple values')
+
+class PythonArgumentsInLuaMethodsTest(PythonArgumentsInLuaTest):
+
+    def __init__(self, *args, **kwargs):
+        super(PythonArgumentsInLuaTest, self).__init__(*args, **kwargs)
```
scoder, line 2718: This is the kind of unexpected hack that requires looking twice to see it. Most people won't.

scoder (commented):
Sorry for the long list of comments. Many of them are style related.

lupa/_lupa.pyx
```diff
@@ -102,6 +102,11 @@ class LuaSyntaxError(LuaError):
     """
 
 
+class LuaMemoryError(LuaError):
+    """Memory error in Lua code.
+    """
```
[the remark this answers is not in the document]
guidanoli: To help distinguish Python and Lua memory errors better?

scoder:
Is that important?

lupa/_lupa.pyx
```diff
@@ -535,6 +547,89 @@ def unpacks_lua_table_method(meth):
...
+    cdef size_t length
+    check_lua_stack(L, 1)
+    lua.lua_pushvalue(L, index)                        # value
+    lua.lua_pushcclosure(L, get_object_length_aux, 1)  # closure
+    result = lua.lua_pcall(L, 0, 1, 0)
+    if result:
+                                                       # err
+        raise_lua_error(runtime, L, result)            #
+                                                       # length
+    length = <size_t>lua.lua_touserdata(L, -1)
```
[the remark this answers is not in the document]
guidanoli: Yes, they refer to the Lua stack. I always put one when the stack is changed or when we know something more about the stack. Maybe the confusion is on `lua_pcall`. We don't know until we evaluate the `result` variable what was pushed onto the stack: an error message, or the return values of the function just called. That's why I put `# err` before `raise_lua_error`, because it is not it who pushed `err`.

lupa/_lupa.pyx
```diff
@@ -535,6 +547,89 @@ def unpacks_lua_table_method(meth):
...
+    cdef lua_State* L = runtime._state
+    check_lua_stack(L, 2)
+    lock_runtime(runtime)
+    old_top = lua.lua_gettop(L)
+    try:
+        length = get_object_length(runtime, L, -1)
+        args = cpython.tuple.PyTuple_New(length)
+        lua.lua_pushnil(L)            # nil (first key)
+        while lua.lua_next(L, -2):
+                                      # key value
```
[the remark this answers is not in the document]
guidanoli: As said previously, I create a comment every time the Lua stack has changed or we know something more about it. `lua_next` only pushes `value` when it returns "true", so I put it inside the while loop.

lupa/_lupa.pyx
```diff
@@ -535,6 +547,89 @@ def unpacks_lua_table_method(meth):
...
+            key = py_from_lua(runtime, L, -2)
+            value = py_from_lua(runtime, L, -1)
+            if isinstance(key, (int, long)) and not isinstance(key, bool):
```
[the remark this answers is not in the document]
guidanoli: Seems like a good idea to me!

lupa/_lupa.pyx
guidanoli: I like the second solution best. It sounds more efficient to allocate the whole tuple beforehand and then shrink it if necessary. I'll look into it.

lupa/tests/test.py
guidanoli: Indeed. It's listed in my first message in this PR.

lupa/tests/test.py
guidanoli: You're right. I think we can get rid of this hack by making `get_args`, `get_kwargs` and `get_none` static methods, like you said.

lupa/tests/test.py
guidanoli: It should. Thanks!

lupa/_lupa.pyx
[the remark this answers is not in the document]
guidanoli: > Is that important?

Clarity of the nature of an error is arguably important for some.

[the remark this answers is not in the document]
guidanoli: But I understand that the memory space of Python and Lua are not much different.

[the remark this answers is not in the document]
scoder: Especially since, under tight memory conditions, even allocating the `LuaMemoryError` instance may fail, in which case you would get a (pre-allocated) Python `MemoryError` instead. :)

lupa/tests/test.py
```diff
@@ -2915,7 +2937,6 @@ def test_functions(self):
         self.testmissingref({}, enumerate)          # enumerate
         self.testmissingref({}, lupa.as_itemgetter) # item getter protocol
         self.testmissingref({}, lupa.as_attrgetter) # attribute getter protocol
-
```
scoder, line 2940: PEP-8 says, two lines of space between top-level definitions.
```suggestion


```

CHANGES.rst
```diff
@@ -10,6 +10,8 @@ Upcoming release
 * The Lua implementation name and version string is provided as
   ``LuaRuntime.lua_implementation``.
 
+* GH#177: Tables that are not sequences raise ``IndexError`` when unpacked.
```
scoder, line 13:
```suggestion
* GH#177: A new function ``python.args(*args, **kwargs)`` was added to help with building
  Python argument tuples and keyword argument dicts for Python function calls from Lua code.

* GH#177: Tables that are not sequences raise ``IndexError`` when unpacked.
  They previously raised a ``TypeError``.
```

scoder:
Nice!

CHANGES.rst
guidanoli: Actually, tables didn't raise any error at all when tables weren't sequences.
Reproducing with older versions:
```python
import lupa
lua = lupa.LuaRuntime()

@lupa.unpacks_lua_table
def foo(*args, **kwargs):
     return args, kwargs

lua.eval('function(f) return f{1, 2, 3, [7] = 0} end')(foo) # -> ((1, 2, 3), {})
```

PR #178, opened 2021-04-17
Export limits for Lua integers

guidanoli:
* Add `lupa.LUA_MININTEGER` and `lupa.LUA_MAXINTEGER`
* Fix integer overflow tests for Windows x86

### DIFF
--- a/lupa/_lupa.pyx
+++ b/lupa/_lupa.pyx
@@ -91,7 +91,18 @@ include "lock.pxi"
 
 cdef int _LUA_VERSION = lua.read_lua_version(NULL)
 LUA_VERSION = (_LUA_VERSION // 100, _LUA_VERSION % 100)
-LUA_MININTEGER, LUA_MAXINTEGER = (lua.LUA_MININTEGER, lua.LUA_MAXINTEGER)
+
+cdef extern from *:
+    const int INT_MIN, INT_MAX
+    const long LONG_MIN, LONG_MAX
+    const long long PY_LLONG_MIN, PY_LLONG_MAX
+
+LUA_MININTEGER, LUA_MAXINTEGER = (
+    (lua.LUA_MININTEGER, lua.LUA_MAXINTEGER) if lua.LUA_VERSION_NUM >= 504 else
+    (PY_LLONG_MIN, PY_LLONG_MAX) if sizeof(lua.lua_Integer) >= sizeof(long long) else  # probably not going to be larger
+    (LONG_MIN, LONG_MAX) if sizeof(lua.lua_Integer) == sizeof(long) else
+    (INT_MIN, INT_MAX)  # probably not going to be smaller
+)
 
 
 class LuaError(Exception):
--- a/lupa/_lupa.pyx
+++ b/lupa/_lupa.pyx
@@ -37,8 +37,10 @@ cdef extern from *:
     #endif
     """
     ctypedef size_t uintptr_t
-    cdef Py_ssize_t PY_SSIZE_T_MAX
-    cdef long LONG_MIN, LONG_MAX
+    cdef const Py_ssize_t PY_SSIZE_T_MAX
+    cdef const int INT_MIN, INT_MAX
+    cdef const long LONG_MIN, LONG_MAX
+    cdef const long long PY_LLONG_MIN, PY_LLONG_MAX
 
 cdef object exc_info
 from sys import exc_info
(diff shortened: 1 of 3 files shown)


lupa/lua.pxd
```diff
@@ -455,3 +455,16 @@ cdef extern from *:
     """
     int read_lua_version(lua_State *L)
     int lua_isinteger(lua_State *L, int idx)
+
+
+cdef extern from "stdint.h":
```
scoder, line 460: Sadly, `stdint.h` is missing from certain older MSVC versions, apparently before VS 2010. And it's a C99 thing. Let's try not to rely on it. Why is `PTRDIFF` the right fallback here?

guidanoli:
lua_Integer is ptrdiff_t in Lua < 5.3
________________________________

lupa/_lupa.pyx
```diff
@@ -90,6 +91,8 @@ include "lock.pxi"
 
 cdef int _LUA_VERSION = lua.read_lua_version(NULL)
 LUA_VERSION = (_LUA_VERSION // 100, _LUA_VERSION % 100)
+LUA_MININTEGER = lua.LUA_MININTEGER
+LUA_MAXINTEGER = lua.LUA_MAXINTEGER
```
scoder, line 95: I think it's ok to do this ourselves, e.g.
```suggestion

cdef extern from *:
    const int INT_MIN, INT_MAX
    const long LONG_MIN, LONG_MAX
    const long long PY_LLONG_MIN, PY_LLONG_MAX

LUA_MININTEGER, LUA_MAXINTEGER = (
    (lua.LUA_MININTEGER, lua.LUA_MAXINTEGER) if lua.LUA_VERSION_NUM >= 504 else
    (PY_LLONG_MIN, PY_LLONG_MAX) if sizeof(lua.lua_Integer) >= sizeof(long long) else  # probably not going to be larger
    (LONG_MIN, LONG_MAX) if sizeof(lua.lua_Integer) == sizeof(long) else
    (INT_MIN, INT_MAX)  # probably not going to be smaller
```

lupa/lua.pxd
```diff
@@ -455,3 +455,16 @@ cdef extern from *:
...
+    # Limits for Lua integers
+    """
+    #if LUA_VERSION_NUM < 503
+    #define LUA_MAXINTEGER PTRDIFF_MAX
+    #define LUA_MININTEGER PTRDIFF_MIN
+    #endif
+    """
```
scoder, line 467:
```suggestion
cdef extern from *:
    # Limits for Lua integers (in Lua<5.4: PTRDIFF_MIN, PTRDIFF_MAX)
    """
    #if LUA_VERSION_NUM < 503
    #define LUA_MAXINTEGER 0
    #define LUA_MININTEGER 0
    #endif
    """
```

lupa/lua.pxd
guidanoli: `lua_Integer` is `ptrdiff_t` in versions of Lua prior to 5.3

guidanoli (commented):
Some issues are to be fixed

lupa/_lupa.pyx
```diff
@@ -92,6 +96,18 @@ cdef int _LUA_VERSION = lua.read_lua_version(NULL)
...
 
+if lua.LUA_VERSION_NUM >= 503:
+    LUA_MININTEGER, LUA_MAXINTEGER = (lua.LUA_MININTEGER, lua.LUA_MAXINTEGER)
+elif sizeof(lua.lua_Integer) >= sizeof(long long):  # probably not larger
+    LUA_MININTEGER, LUA_MAXINTEGER = (PY_LLONG_MIN, PY_LLONG_MAX)
+elif sizeof(lua.lua_Integer) >= sizeof(long):
+    LUA_MININTEGER, LUA_MAXINTEGER = (LONG_MIN, LONG_MAX)
+elif sizeof(lua.lua_Integer) >= sizeof(int):
+    LUA_MININTEGER, LUA_MAXINTEGER = (INT_MIN, INT_MAX)
+else:  # probably not smaller
```
scoder, line 107: What about `short`? I mean, I wouldn't expect anyone to define `lua_Integer` as anything smaller than `long`, let alone `int`, but if we call out `char`, then I'd also expect `short int` on that list. (Mind you, I can live without it – I just find the jump from `int` to `char` striking.)

guidanoli: You're right.

scoder:
Perfect. Fixes the failing tests on appveyor. Thanks!

PR #181, opened 2021-04-26
Make python.enumerate yield integer indices

### LINKED ISSUE #180: python.enumerate yields indices in floating-point format
**Short description:**
The current implementation of `python.enumerate` uses `lua_Number` as indices, while we could use `lua_Integer` for greater range and precision

**Reasoning:**
Although `lua_Integer` and `lua_Number` have the same size, their ranges are different. Specifically, `lua_Number` reserves some of its bits for the mantissa, while all but one bit from `lua_Integer` serve that purpose. Since floating point numbers don't really make sense in enumeration, we should be using integers as indices.

**How:**
Code relating to `python.enumerate` should use:
* `lua_Integer` instead of `lua_Number`
* `1` instead of `1.0`
* `0` instead of `0.0`
* `lua_pushinteger` instead of `lua_pushnumber` (1)
* `lua_tointeger` instead of `lua_tonumber` (1)

_Note (1): These functions were present in the Lua API prior to 5.3 for convenience. Back then, integers were converted to `lua_Number` under the table._

**Effect:**
For versions of Lua prior to 5.3, this won't make a difference, since all numbers are represented in floating-point format.
But for versions of Lua later than 5.3, when integers were introduced, this will expand the range of enumerated iterators.

guidanoli:
Closes #180

### DIFF
--- a/lupa/_lupa.pyx
+++ b/lupa/_lupa.pyx
@@ -1904,11 +1904,27 @@ cdef int py_iterex(lua_State* L) nogil:
         return lua.lua_error(L)  # never returns!
     return result
 
+cdef int convert_to_lua_Integer(lua_State* L, int idx, lua.lua_Integer* integer) nogil:
+    cdef int isnum
+    cdef lua.lua_Integer temp
+    temp = lua.lua_tointegerx(L, idx, &isnum)
+    if isnum:
+        integer[0] = temp
+        return 0
+    else:
+        lua.lua_pushfstring(L, "Could not convert %s to string", lua.luaL_typename(L, idx))
+        return -1
+
 cdef int py_enumerate(lua_State* L) nogil:
     if lua.lua_gettop(L) > 2:
         lua.luaL_argerror(L, 3, "invalid arguments")   # never returns!
     cdef py_object* py_obj = unpack_python_argument_or_jump(L, 1)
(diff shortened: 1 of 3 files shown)


lupa/_lupa.pyx
```diff
@@ -1908,19 +1908,19 @@ cdef int py_enumerate(lua_State* L) nogil:
     if lua.lua_gettop(L) > 2:
         lua.luaL_argerror(L, 3, "invalid arguments")   # never returns!
     cdef py_object* py_obj = unpack_python_argument_or_jump(L, 1)
-    cdef double start = lua.lua_tonumber(L, -1) if lua.lua_gettop(L) == 2 else 0.0
+    cdef lua.lua_Integer start = lua.lua_tointeger(L, -1) if lua.lua_gettop(L) == 2 else 0
```
scoder, line 1911: Lua 5.2+ has `lua_tointegerx()`, which reports success or failure with an additional return argument. It seems worth wrapping that in a helper function to raise appropriate exceptions for overflows etc.
A fallback for 5.1 could use a C macro to map it to `lua_tointeger()` and set the error argument to 1 (i.e. it was a number).

guidanoli: Thanks!

scoder:
Thanks

PR #183, opened 2021-04-29
Fix python.args usage and optimize call_python

### LINKED ISSUE #182: The usage of python.args is inconsistent with Lua varargs semantics
**Introduction:**
With the recent introduction of `python.args`, a subtle semantic defect relating to Lua varargs has leaked.
Moreover, the performance of Python calls in Lua (with and without `python.args`) would be greatly improved as a side effect of a more restricted usage of `python.args`.

**Reasoning:**
Currently, in Lua, the ellipsis token (`...`) is used for representing varargs inside functions.

```lua
-- talk: example of varargs in functions
function talk(name, ...)
   print(name, 'says', ...)
end

talk("Bob", "the answer is", 6 * 7)
-- prints "Bob says the answer is 42"

talk("Bob")
-- prints "Bob says"
```

Additionally, the `...` expression can be used in table constructors.

```lua
-- pack: example of varargs in table constructors
function pack(...) return {...} end

t = pack(1, 2, 3)
for _, v in ipairs(t) do
   print(v)
end
-- prints 1 2 3
```

However, with multiple varargs, **only the last one is expanded**.

```lua
-- pack3: example of multiple varargs
function pack3(...) return {..., ..., ...} end

t = pack3(1, 2, 3)
for _, v in ipairs(t) do
   print(v)
end
-- prints 1 1 1 2 3
```

Analogously, function calls also have their return values expanded.

```lua
-- get123: example of multiple return values
function get123() return 1, 2, 3

print(get123())
-- prints 1 2 3
```

However, with multiple function calls, **only the last one is expanded**.

```lua
print(get123(), get123(), get123())
-- prints 1 1 1 2 3
```

In contrast, the current handling of `python.args` allows for multiple instances in the same call.

```lua
local get123 = python.args{1, 2, 3}
python.builtins.print(get123, get123, get123)
-- prints 1 2 3 1 2 3 1 2 3
```

Notice the dissonance between this example and the previous, in pure Lua.

**Proposed solutions:**

One option would be to constrain the usage `python.args` as the last argument on a function call.
This is the solution most cohesive with the current Lua varargs semantics, since `python.args` would **only be expanded if last**.
As a side bonus, detecting and handling `python.args` would have significant performance gains.

A second option would be to restrict the usage of `python.args` for keyword arguments only.
This would not clash with current Lua semantics, since it does not support keyword arguments.
Note that table unpacking is possible anyways with the [`table.unpack`](https://www.lua.org/manual/5.4/manual.html#6.6) function in standard Lua.
Furthermore, this would also bring significant performance benefits to the current implementation of Python calls in Lua.

guidanoli:
Closes #182

### DIFF
--- a/lupa/_lupa.pyx
+++ b/lupa/_lupa.pyx
@@ -1648,6 +1648,7 @@ cdef int py_object_gc(lua_State* L) nogil:
 cdef bint call_python(LuaRuntime runtime, lua_State *L, py_object* py_obj) except -1:
     # Callers must assure that py_obj.obj is not NULL, i.e. it points to a valid Python object.
     cdef int i, nargs = lua.lua_gettop(L) - 1
+    cdef Py_ssize_t j
     cdef tuple args
     cdef dict kwargs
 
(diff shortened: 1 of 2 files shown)


scoder:
You say that this optimises the call. So, how much faster is it?

lupa/_lupa.pyx
```diff
@@ -1657,16 +1658,31 @@ cdef bint call_python(LuaRuntime runtime, lua_State *L, py_object* py_obj) excep
...
-                args += (arg, )
+        # Special treatment for the last argument
+        last_arg = py_from_lua(runtime, L, nargs + 1)
+
+        if isinstance(last_arg, _PyArguments):
+            # Calling a function and _PyArguments is the last argument
+            # Lua f(..., python.args{a, b, c=1, d=2}) => Python as f(..., a, b, c=1, d=2)
+            kwargs = (<_PyArguments>last_arg).kwargs
+            moreargs = (<_PyArguments>last_arg).args
+            args = cpython.tuple.PyTuple_New(nargs - 1 + cpython.tuple.PyTuple_Size(moreargs))
```
scoder, line 1669: Just use `len()`, Cython knows that this is a tuple.
```suggestion
            moreargs = (<_PyArguments>last_arg).args
            args = cpython.tuple.PyTuple_New(nargs - 1 + len(moreargs))
```

guidanoli:
> You say that this optimises the call. So, how much faster is it?

It's trivial to understand that this is faster. There's no need for benchmarking it.

*Before:*
* has to check every argument to see it is a _PyArguments instance (O(n))
* has to merge O(n) dictionaries
* has to concatenate O(n) tuples

*After:*
* has to check only the last argument (O(1))
* there is either 1 dictionary or none
* pre-allocates 1 tuple of right size for all positional arguments

guidanoli:
See #187 for updated changes

PR #188, opened 2021-07-16
Safer handling of the Lua stack

guidanoli:
Some places don't lock the LuaRuntime before fiddling with the Lua stack, and many places don't ensure extra slots either.
This PR also involves other changes:
* Add preconditions and postconditions to important functions/methods
* Change behaviour `_LuaObject.__str__`
  * If `__tostring` doesn't exist, use `_LuaObject.__repr__`
  * If `__tostring` returns non-string value, raise a `TypeError`
  * If `__tostring` raises an error, raise a `LuaError`
* `lua_pushglobaltable` instead of pushing `_G`, which may not point to the global table.
* Make `lock_runtime` return boolean and not raise error directly
* Use `LUA_NOREF` and `LUA_REFNIL` constants for effective handling of Lua references
* Add important assertions (checks if `LuaRuntime._state` is not `NULL`, and if `_LuaObject._runtime` is `None`, etc...)

### DIFF
--- a/lupa/_lupa.pyx
+++ b/lupa/_lupa.pyx
@@ -132,6 +132,7 @@ def lua_type(obj):
         return None
     lua_object = <_LuaObject>obj
     assert lua_object._runtime is not None
+    assert lua_object._runtime._state is not NULL
     lock_runtime(lua_object._runtime)
     L = lua_object._state
     old_top = lua.lua_gettop(L)
@@ -291,6 +292,7 @@ cdef class LuaRuntime:
         """
         The Lua runtime/language version as tuple, e.g. (5, 3) for Lua 5.3.
         """
+        assert self._state is not NULL
         cdef int version = lua.read_lua_version(self._state)
         return (version // 100, version % 100)
 
@@ -343,6 +345,7 @@ cdef class LuaRuntime:
     def compile(self, lua_code):
         """Compile a Lua program into a callable Lua function.
         """
+        assert self._state is not NULL
         cdef const char *err
         if isinstance(lua_code, unicode):
             lua_code = (<unicode>lua_code).encode(self._source_encoding)
@@ -460,6 +463,7 @@ cdef class LuaRuntime:
     def set_overflow_handler(self, overflow_handler):
         """Set the overflow handler function that is called on failures to pass large numbers to Lua.
         """
+        assert self._state is not NULL
         cdef lua_State *L = self._state
         if overflow_handler is not None and not callable(overflow_handler):
             raise ValueError("overflow_handler must be callable")
@@ -670,6 +674,8 @@ cdef tuple _fix_args_kwargs(tuple args):
         return args, {}
 
     cdef _LuaTable table = <_LuaTable>arg
+    assert table._runtime is not None
+    assert table._runtime._state is not NULL
     cdef LuaRuntime runtime = table._runtime
     cdef lua_State* L = table._state
     lock_runtime(runtime)
@@ -706,6 +712,9 @@ cdef class _LuaObject:
     cdef lua_State* _state
     cdef int _ref
 
+    def __cinit__(self):
+        self._ref = lua.LUA_NOREF
+
     def __init__(self):
         raise TypeError("Type cannot be instantiated manually")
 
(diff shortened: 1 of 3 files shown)


guidanoli:
Can you review this, @scoder ?

scoder:
Did you test if this has any performance impact? (I.e., can you prove that it doesn't?)

lupa/_lupa.pyx
```diff
@@ -150,9 +148,58 @@ def lua_type(obj):
...
+    """Context handler for the Lua runtime stack
+    Invariants:
+        _runtime is not None
+    """
+    cdef LuaRuntime _runtime
+    cdef int _top
+    cdef int _extra
+    cdef int _restore
+
+    def __enter__(self):
```
scoder, line 172: You can make this a `cdef` method.

lupa/_lupa.pyx
```diff
@@ -150,9 +148,58 @@ def lua_type(obj):
...
+
+
+cdef enum _LuaRuntimeStackRestorationPolicy:
+    # restoration policy for Lua stack context handler
+    RESTORE_NEVER = 0
+    RESTORE_ALWAYS = 1
+    RESTORE_ON_ERROR = 2
+
+@cython.internal
+@cython.no_gc_clear
```
scoder, line 160: Why is `no_gc_clear` needed here?

lupa/_lupa.pyx
```diff
@@ -654,17 +683,16 @@ cdef tuple _fix_args_kwargs(tuple args):
...
 
 ################################################################################
 # fast, re-entrant runtime locking
 
-cdef inline int lock_runtime(LuaRuntime runtime) except -1:
-    if not lock_lock(runtime._lock, pythread.PyThread_get_thread_ident(), True):
-        raise LuaError("Failed to acquire thread lock")
-    return 0
+cdef inline bint lock_runtime(LuaRuntime runtime) with gil:
+    return lock_lock(runtime._lock, pythread.PyThread_get_thread_ident(), True)
```
scoder, line 695: This used to raise an exception and it no longer does. Is that intended?

lupa/_lupa.pyx
```diff
@@ -1554,15 +1532,22 @@ cdef object execute_lua_call(LuaRuntime runtime, lua_State *L, Py_ssize_t nargs)
...
     if args:
-        old_top = lua.lua_gettop(L)
-        for i, arg in enumerate(args):
-            if not py_to_lua(runtime, L, arg, wrap_none=not first_may_be_nil):
-                lua.lua_settop(L, old_top)
-                raise TypeError("failed to convert argument at index %d" % i)
-            first_may_be_nil = True
-    return 0
+        nargs = len(args)
+        if nargs > INT_MAX:
```
scoder, line 1540: This might give a C compiler warning on systems where `sizeof(int)` == `sizeof(Py_ssize_t)`, since the condition would always be false there.

lupa/_lupa.pyx
scoder: (read your description, it _is_ intended)

Why is the `with gil` declaration necessary?

lupa/_lupa.pyx
guidanoli: Wouldn't it be simply optimized out?

guidanoli:
> Did you test if this has any performance impact? (I.e., can you prove that it doesn't?)

Previously, with the context handler it was 50% the speed.
Now we use inline functions `getstacktop` and `setstacktop`, which is now 94% the speed, which I think it's pretty acceptable since the gains in readability and maintainability are greater.

guidanoli:
Note that `getstack` does more than just lock the runtime, check the stack and get the stack top. It also unlocks the runtime if `check_lua_stack` raises an error, which is quite important. Otherwise, this resource would not be properly released.

guidanoli:
I change my mind -- I think it's fine the way it is, explicit and efficient.
However, I still check for extra stack space wherever needed.

guidanoli:
Could you review this, @scoder? It makes Lupa more reliable, in that we guarantee we can push new values to the Lua stack.

lupa/_lupa.pyx
```diff
@@ -1327,26 +1383,43 @@ cdef int py_function_result_to_lua(LuaRuntime runtime, lua_State *L, object o) e
...
+    check_lua_stack(L, 2)
+    old_top = lua.lua_gettop(L)
+    try:
+        lua.lua_pushlstring(L, LUPAOFH, len(LUPAOFH))
+        lua.lua_rawget(L, lua.LUA_REGISTRYINDEX)
+        if lua.lua_isnil(L, -1):
+            lua.lua_pop(L, 1)
+            return 0
+        py_to_lua_custom(runtime, L, o, 0)
+        if lua.lua_pcall(L, 1, 1, 0):
```
scoder, line 1406: Are you sure that this can never put more than one item on the stack? I think that used to be possible.

lupa/_lupa.pyx
scoder: Probably, yes, after giving the warning.

scoder:
Yes, this looks very clean and helpful.

lupa/_lupa.pyx
guidanoli: Although the current implementation of `py_to_lua_custom` always pushes 1 item, I agree that we should make it generic.

guidanoli: Actually, the postcondition for `py_to_lua_custom` is that it pushes one value to the Lua stack and that it returns 1 (except errors).

guidanoli:
Let's get this merged? I depend on it for future PRs.

scoder:
Works for me. Thanks!

Issue #196, opened 2021-10-31
Provide all of Lua 5.3/5.4 and LuaJIT in a single package


romanalexander:
The distributions from pip use the bundled Lua 5.3 implementation, not LuaJIT2.0.5 as the readme suggests

scoder:
Hmm. Given how small the Lua implementations are, I think it would be cool if the Lupa wheels shipped just all available options, say, Lua 5.3/4 and LuaJIT, as a Package. Then importing "lupa" would just give you the latest Lua 5.x, as it's the case now, and importing "lupa.luajit2" would give you the same API but built with LuaJIT 2.x. This could all come from the current module implementation with a more elaborate build setup.

PR very welcome.

scoder:
Module names to provide:
- `lupa` - `from .lua54 import *` (i.e. delegating to latest Lua module)
- `lupa.lua53` - Lua 5.3.x
- `lupa.lua54` - Lua 5.4.x
- `lupa.luajit20` - LuaJIT 2.0.x
- `lupa.luajit21` - LuaJIT 2.1.x

tatuylonen:
I strongly support making lupa available with different Lua versions!  I'm using lupa extensively for Wiktionary extraction (https://github.com/tatuylonen/wiktextract, or actually its subpackage https://github.com/tatuylonen/wikitextprocessor).  Wiktionary uses Lua 5.2, and I have no way of upgrading the Wiktionary modules to Lua 5.3 or 5.4.  I've coped with Lua 5.3 by patching the Lua source code using essentially a few dozen regexp-replaces, but it is a horrible kludge.  Now with Lua 5.4, it has again changed in several incompatible ways, including changes in supported format strings and apparently the operation of metatables has also changed in subtle ways.  It's coming to a point where I'll have to start building and releasing my own version of lupa compiled for an older version unless there is support for either selecting the version to use when creating a Lua context or installing a separate package.  The proposed way of, e.g., lupa.lua53 works, but I would very much prefer using Lua 5.2.  The English Wiktionary alone contains 322,000 lines of Lua 5.2 code, and when counting the other language editions and Wikipedia, it is probably millions of lines of Lua 5.2 code in Mediawiki projects alone.  It is not going to get upgraded from Lua 5.2 any time soon, I think.

I came to this repository with the intention of submitting an issue to request this feature.

scoder:
See #207

scoder:
[...]

PR #207, opened 2022-03-06
Ship multiple Lua versions

scoder:
Currently lacks support for building the bundled LuaJIT on Windows (and macOS, but that's a different story).
Also, building Lua 5.2.3 currently fails and thus isn't integrated.

Closes https://github.com/scoder/lupa/issues/196

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -60,7 +60,7 @@ jobs:
       run: sudo apt-get install lib${{ matrix.lua-version }}-dev
 
     - name: Build wheel
-      run: python -m pip install -r requirements.txt && python setup.py sdist build_ext -j5 bdist_wheel
+      run: python -m pip install -r requirements.txt && python setup.py sdist ${{ contains(matrix.python-version, '3.') && 'build_ext -j5' || '' }} bdist_wheel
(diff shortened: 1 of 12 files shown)


tatuylonen:
I'm trying to build the multilua branch on Linux Mint 20.2 (essentially Ubuntu; Python 3.8.10) with "python3 setup.py build".  It fails with "FileNotFoundError: [Errno 2] No such file or directory: '/home/ylo/lupa/orig/lupa/third-party/lua53/makefile'.  Apparently the third party modules need to be installed from somewhere.  It is not immediately clear to me from where to get the correct versions.  Is that mentioned somewhere or is there a script to fetch them?

At least on Linux I think it would be fine to just use the precompiled liblua-*-dev packages and copy libraries from them (as with --lua-lib --lua-includes --use-bundle before).  At least Ubuntu has liblua5.2-dev.

scoder:
> `FileNotFoundError: [Errno 2] No such file or directory: '/home/ylo/lupa/orig/lupa/third-party/lua53/makefile'`

You probably just need to check out the git submodules. `git submodule update --init --recursive`. I guess that should be spelled out in the README somewhere…

Issue #199, opened 2021-12-13
nested dict to lua dictionary


jappoz:
Hello, 
Is there any way to use table_from method given a nested dictionary to produce the corresponding lua table?
What I would like to achieve is something like this:

```
#python
main_dict = {}
main_dict_core = {}
main_dict_core['A'] = True
main_dict_core['B'] = [{4,6}, {7.7, 8}]
main_dict['MAIN'] = main_dict_core
out_table = lua.table_from(main_dict)

# lua
MAIN = {
  A = true,
  B = {
    {4, 6},
    {7.7, 8}
  },
}
```
and then convert out_table to string and save it as lua file.

I have seen in #34 that a recursive=True arg could be used but it does not seem to be implemented so far...

scoder:
> I have seen in #34 that a recursive=True arg could be used but it does not seem to be implemented so far...

Seems worth implementing then. PR welcome.

See
https://github.com/scoder/lupa/blob/7fe8768bd2bc3fbba776099a384519a549972cb2/lupa/_lupa.pyx#L430

The tests are in lupa/tests/test.py, Look for "table_from".

synodriver:
Wait https://github.com/scoder/lupa/pull/208

kebab-mai-haddi:
Is there a way to convert a Lua table with circular references into a Python dictionary?

synodriver:
> Is there a way to convert a Lua table with circular references into a Python dictionary?

Detecting circular references data structures is really annoying in this case, we'd better limit the depth in some way.

PR #208, opened 2022-04-02
add nested dict support

synodriver:
See [this issue](https://github.com/scoder/lupa/issues/199)

### DIFF
--- a/lupa/_lupa.pyx
+++ b/lupa/_lupa.pyx
@@ -48,10 +48,11 @@ cdef object exc_info
 from sys import exc_info
 
 cdef object Mapping
+cdef object Sequence
 try:
-    from collections.abc import Mapping
+    from collections.abc import Mapping, Sequence
 except ImportError:
-    from collections import Mapping  # Py2
+    from collections import Mapping, Sequence  # Py2
 
 cdef object wraps
 from functools import wraps
@@ -408,7 +409,7 @@ cdef class LuaRuntime:
         """
         return self.table_from(items, kwargs)
 
-    def table_from(self, *args):
+    def table_from(self, *args, bint recursive=False):
         """Create a new table from Python mapping or iterable.
 
         table_from() accepts either a dict/mapping or an iterable with items.
@@ -416,7 +417,7 @@ cdef class LuaRuntime:
         are placed in the table in order.
 
         Nested mappings / iterables are passed to Lua as userdata
-        (wrapped Python objects); they are not converted to Lua tables.
+        (wrapped Python objects) if recursive is False; they are not converted to Lua tables.
         """
         assert self._state is not NULL
         cdef lua_State *L = self._state
@@ -426,12 +427,12 @@ cdef class LuaRuntime:
         try:
             check_lua_stack(L, 5)
             lua.lua_newtable(L)
-            # FIXME: how to check for failure?
+            # FIXME: how to check for failure? and nested dict
             for obj in args:
                 if isinstance(obj, dict):
-                    for key, value in obj.iteritems():
-                        py_to_lua(self, L, key)
-                        py_to_lua(self, L, value)
+                    for key, value in obj.iteritems():  # in python3, this is called items
+                        py_to_lua(self, L, key, False, recursive)
+                        py_to_lua(self, L, value, False, recursive)
                         lua.lua_rawset(L, -3)
 
                 elif isinstance(obj, _LuaTable):
@@ -447,12 +448,12 @@ cdef class LuaRuntime:
                 elif isinstance(obj, Mapping):
                     for key in obj:
                         value = obj[key]
-                        py_to_lua(self, L, key)
-                        py_to_lua(self, L, value)
+                        py_to_lua(self, L, key, False, recursive)
+                        py_to_lua(self, L, value, False, recursive)
                         lua.lua_rawset(L, -3)
                 else:
                     for arg in obj:
-                        py_to_lua(self, L, arg)
+                        py_to_lua(self, L, arg, False, recursive)
                         lua.lua_rawseti(L, -2, i)
                         i += 1
             return py_from_lua(self, L, -1)
@@ -508,12 +509,12 @@ cdef class LuaRuntime:
         luaL_openlib(L, "python", py_lib, 0)       # lib
         lua.lua_pushlightuserdata(L, <void*>self)  # lib udata
         lua.lua_pushcclosure(L, py_args, 1)        # lib function
-        lua.lua_setfield(L, -2, "args")            # lib 
+        lua.lua_setfield(L, -2, "args")            # lib
 
         # register our own object metatable
         lua.luaL_newmetatable(L, POBJECT)          # lib metatbl
         luaL_openlib(L, NULL, py_object_lib, 0)
-        lua.lua_pop(L, 1)                          # lib 
+        lua.lua_pop(L, 1)                          # lib
 
         # create and store the python references table
         lua.lua_newtable(L)                                  # lib tbl
@@ -1135,7 +1136,7 @@ cdef object resume_lua_thread(_LuaThread thread, tuple args):
             # already terminated
             raise StopIteration
         if args:
-            nargs = len(args)
+            nargs = <int>len(args)
             push_lua_arguments(thread._runtime, co, args)
         with nogil:
             status = lua.lua_resume(co, L, nargs, &nres)
@@ -1381,7 +1382,7 @@ cdef py_object* unpack_userdata(lua_State *L, int n) nogil:
 cdef int py_function_result_to_lua(LuaRuntime runtime, lua_State *L, object o) except -1:
      if runtime._unpack_returned_tuples and isinstance(o, tuple):
          push_lua_arguments(runtime, L, <tuple>o)
-         return len(<tuple>o)
+         return <int>len(<tuple>o)
      check_lua_stack(L, 1)
      return py_to_lua(runtime, L, o)
 
@@ -1410,7 +1411,7 @@ cdef int py_to_lua_handle_overflow(LuaRuntime runtime, lua_State *L, object o) e
         lua.lua_settop(L, old_top)
         raise
 
-cdef int py_to_lua(LuaRuntime runtime, lua_State *L, object o, bint wrap_none=False) except -1:
+cdef int py_to_lua(LuaRuntime runtime, lua_State *L, object o, bint wrap_none=False, bint recursive=False) except -1:
     """Converts Python object to Lua
     Preconditions:
         1 extra slot in the Lua stack
@@ -1461,6 +1462,19 @@ cdef int py_to_lua(LuaRuntime runtime, lua_State *L, object o, bint wrap_none=Fa
         pushed_values_count = 1
     elif isinstance(o, float):
         lua.lua_pushnumber(L, <lua.lua_Number><double>o)
+        pushed_values_count = 1
+    elif recursive and isinstance(o, Sequence):
+        lua.lua_createtable(L, <int>len(o), 0)   # create a table at the top of stack, with narr already known
+        for i, v in enumerate(o):
+            py_to_lua(runtime, L, v, wrap_none, recursive)
+            lua.lua_rawseti(L, -2, i+1)
+        pushed_values_count = 1
+    elif recursive and isinstance(o, Mapping):
+        lua.lua_createtable(L, 0, <int>len(o))  # create a table at the top of stack, with nrec already known
+        for key, value in o.iteritems(): # to compatible with py2
+            py_to_lua(runtime, L, key, wrap_none, recursive)
+            py_to_lua(runtime, L, value, wrap_none, recursive)
+            lua.lua_rawset(L, -3)
         pushed_values_count = 1
     else:
         if isinstance(o, _PyProtocolWrapper):
@@ -1626,7 +1640,7 @@ cdef object execute_lua_call(LuaRuntime runtime, lua_State *L, Py_ssize_t nargs)
                 lua.lua_replace(L, -2)
                 lua.lua_insert(L, 1)
                 errfunc = 1
-        result_status = lua.lua_pcall(L, nargs, lua.LUA_MULTRET, errfunc)
+        result_status = lua.lua_pcall(L, <int>nargs, lua.LUA_MULTRET, errfunc)
         if errfunc:
             lua.lua_remove(L, 1)
     results = unpack_lua_results(runtime, L)
@@ -1736,7 +1750,7 @@ cdef int py_object_gc_with_gil(py_object *py_obj, lua_State* L) with gil:
         return 0
     finally:
         py_obj.obj = NULL
-    
+
 cdef int py_object_gc(lua_State* L) nogil:
     if not lua.lua_isuserdata(L, 1):
         return 0
@@ -1762,7 +1776,6 @@ cdef bint call_python(LuaRuntime runtime, lua_State *L, py_object* py_obj) excep
     else:
         args = ()
         kwargs = {}
-        
         for i in range(nargs):
             arg = py_from_lua(runtime, L, i+2)
             if isinstance(arg, _PyArguments):
--- a/lupa/tests/test.py
+++ b/lupa/tests/test.py
@@ -24,6 +24,7 @@ unicode_type = type('abc'.decode('ASCII') if IS_PYTHON2 else 'abc')
 if IS_PYTHON2:
     unittest.TestCase.assertRaisesRegex = unittest.TestCase.assertRaisesRegexp
 
+
 class SetupLuaRuntimeMixin(object):
     lua_runtime_kwargs = {}
 
@@ -97,7 +98,7 @@ class TestLuaRuntime(SetupLuaRuntimeMixin, unittest.TestCase):
         self.assertEqual(2, self.lua.eval('1+1'))
 
     def test_eval_multi(self):
-        self.assertEqual((1,2,3), self.lua.eval('1,2,3'))
+        self.assertEqual((1, 2, 3), self.lua.eval('1,2,3'))
 
     def test_eval_args(self):
         self.assertEqual(2, self.lua.eval('...', 2))
@@ -170,7 +171,7 @@ class TestLuaRuntime(SetupLuaRuntimeMixin, unittest.TestCase):
         return fac
         ''')
         self.assertNotEqual(None, fac)
-        self.assertEqual(6,       fac(3))
+        self.assertEqual(6, fac(3))
         self.assertEqual(3628800, fac(10))
 
     def test_double_recursive_function(self):
@@ -185,8 +186,8 @@ class TestLuaRuntime(SetupLuaRuntimeMixin, unittest.TestCase):
         '''
         calc = self.lua.execute(func_code)
         self.assertNotEqual(None, calc)
-        self.assertEqual(3,     calc(3))
-        self.assertEqual(109,   calc(10))
+        self.assertEqual(3, calc(3))
+        self.assertEqual(109, calc(10))
         self.assertEqual(13529, calc(20))
 
     def test_double_recursive_function_pycallback(self):
@@ -199,14 +200,15 @@ class TestLuaRuntime(SetupLuaRuntimeMixin, unittest.TestCase):
         end
         return calc
         '''
+
         def pycallback(i):
-            return i**2
+            return i ** 2
 
         calc = self.lua.execute(func_code)
 
         self.assertNotEqual(None, calc)
-        self.assertEqual(12,     calc(pycallback, 3))
-        self.assertEqual(1342,   calc(pycallback, 10))
+        self.assertEqual(12, calc(pycallback, 3))
+        self.assertEqual(1342, calc(pycallback, 10))
         self.assertEqual(185925, calc(pycallback, 20))
 
     def test_none(self):
@@ -243,6 +245,7 @@ class TestLuaRuntime(SetupLuaRuntimeMixin, unittest.TestCase):
 
     def test_call_str_class(self):
         called = [False]
+
         class test(object):
             def __str__(self):
                 called[0] = True
@@ -267,7 +270,7 @@ class TestLuaRuntime(SetupLuaRuntimeMixin, unittest.TestCase):
 
     def test_len_table_dict(self):
         table = self.lua.eval('{a=1, b=2, c=3}')
-        self.assertEqual(0, len(table)) # as returned by Lua's "#" operator
+        self.assertEqual(0, len(table))  # as returned by Lua's "#" operator
 
     def test_table_delattr(self):
         table = self.lua.eval('{a=1, b=2, c=3}')
@@ -293,27 +296,27 @@ class TestLuaRuntime(SetupLuaRuntimeMixin, unittest.TestCase):
 
     def test_iter_table(self):
         table = self.lua.eval('{2,3,4,5,6}')
-        self.assertEqual([1,2,3,4,5], list(table))
+        self.assertEqual([1, 2, 3, 4, 5], list(table))
 
     def test_iter_table_list_repeat(self):
         table = self.lua.eval('{2,3,4,5,6}')
-        self.assertEqual([1,2,3,4,5], list(table))  # 1
-        self.assertEqual([1,2,3,4,5], list(table))  # 2
-        self.assertEqual([1,2,3,4,5], list(table))  # 3
+        self.assertEqual([1, 2, 3, 4, 5], list(table))  # 1
+        self.assertEqual([1, 2, 3, 4, 5], list(table))  # 2
+        self.assertEqual([1, 2, 3, 4, 5], list(table))  # 3
 
     def test_iter_array_table_values(self):
         table = self.lua.eval('{2,3,4,5,6}')
-        self.assertEqual([2,3,4,5,6], list(table.values()))
+        self.assertEqual([2, 3, 4, 5, 6], list(table.values()))
 
     def test_iter_array_table_repeat(self):
         table = self.lua.eval('{2,3,4,5,6}')
-        self.assertEqual([2,3,4,5,6], list(table.values()))  # 1
-        self.assertEqual([2,3,4,5,6], list(table.values()))  # 2
-        self.assertEqual([2,3,4,5,6], list(table.values()))  # 3
+        self.assertEqual([2, 3, 4, 5, 6], list(table.values()))  # 1
+        self.assertEqual([2, 3, 4, 5, 6], list(table.values()))  # 2
+        self.assertEqual([2, 3, 4, 5, 6], list(table.values()))  # 3
 
     def test_iter_multiple_tables(self):
         count = 10
-        table_values = [self.lua.eval('{%s}' % ','.join(map(str, range(2, count+2)))).values()
+        table_values = [self.lua.eval('{%s}' % ','.join(map(str, range(2, count + 2)))).values()
                         for _ in range(4)]
 
         # round robin
@@ -322,11 +325,11 @@ class TestLuaRuntime(SetupLuaRuntimeMixin, unittest.TestCase):
             for table in table_values:
                 sublist.append(_next(table))
 
-        self.assertEqual([[i]*len(table_values) for i in range(2, count+2)], l)
+        self.assertEqual([[i] * len(table_values) for i in range(2, count + 2)], l)
 
     def test_iter_table_repeat(self):
         count = 10
-        table_values = [self.lua.eval('{%s}' % ','.join(map(str, range(2, count+2)))).values()
+        table_values = [self.lua.eval('{%s}' % ','.join(map(str, range(2, count + 2)))).values()
                         for _ in range(4)]
 
         # one table after the other
@@ -335,7 +338,7 @@ class TestLuaRuntime(SetupLuaRuntimeMixin, unittest.TestCase):
             for sublist in l:
                 sublist.append(_next(table))
 
-        self.assertEqual([[i]*len(table_values) for i in range(2,count+2)], l)
+        self.assertEqual([[i] * len(table_values) for i in range(2, count + 2)], l)
 
     def test_iter_table_refcounting(self):
         lua_func = self.lua.eval('''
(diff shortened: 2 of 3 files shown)


synodriver:
Besides, I fixed some compiler warnings on msvc about type convertion from ```Py_ssize_t``` to ```int```.

synodriver:
Wow the ci fails somehow. You should build lua first and link extension against lua.dll or liblua.so. See my github action's [script](https://github.com/synodriver/lupa/blob/winbuild/.github/workflows/build_wheel.yml)

lupa/_lupa.pyx
```diff
@@ -1462,6 +1463,19 @@ cdef int py_to_lua(LuaRuntime runtime, lua_State *L, object o, bint wrap_none=Fa
...
+        pushed_values_count = 1
+    elif recursive and isinstance(o, Mapping):
+        lua.lua_createtable(L, 0, <int>len(o))  # create a table at the top of stack, with nrec already known
+        for key, value in o.iteritems(): # to compatible with py2
+            py_to_lua(runtime, L, key, wrap_none, recursive)
+            py_to_lua(runtime, L, value, wrap_none, recursive)
+            lua.lua_rawset(L, -3)
+        pushed_values_count = 1
     else:
         if isinstance(o, _PyProtocolWrapper):
```
scoder, line 1480: We should keep this case early. It's a cheap type check compared to the ABCs `Sequence` and `Mapping`, and not so unlikely to happen.

lupa/_lupa.pyx
```diff
@@ -426,12 +427,12 @@ cdef class LuaRuntime:
...
-            # FIXME: how to check for failure?
+            # FIXME: how to check for failure? and nested dict
             for obj in args:
                 if isinstance(obj, dict):
-                    for key, value in obj.iteritems():
-                        py_to_lua(self, L, key)
-                        py_to_lua(self, L, value)
+                    for key, value in obj.iteritems():  # in python3, this is called items
+                        py_to_lua(self, L, key, False, recursive)
+                        py_to_lua(self, L, value, False, recursive)
```
scoder, line 435: Please use keyword arguments if you pass constants like `True` and `False`, since without knowing which argument they refer to, it's unclear what they mean.

lupa/tests/test.py
```diff
@@ -97,7 +98,7 @@ def test_eval(self):
         self.assertEqual(2, self.lua.eval('1+1'))
 
     def test_eval_multi(self):
-        self.assertEqual((1,2,3), self.lua.eval('1,2,3'))
+        self.assertEqual((1, 2, 3), self.lua.eval('1,2,3'))
```
scoder, line 101: All whitespace changes are unrelated to this PR. It's mostly ok to make them (although it makes the git history more difficult to use), but I'd rather have them in a separate PR.

lupa/tests/test.py
```diff
@@ -591,6 +594,58 @@ def test_table_from_table_iter_indirect(self):
         self.assertEqual(list(table2.keys()), [1, 2, 3])
         self.assertEqual(set(table2.values()), set([1, 2, "foo"]))
 
+    def test_table_from_nested_dict(self):
+        data = {"a": {"a": "foo"}, "b": {"b": "bar"}}
+        table = self.lua.table_from(data, recursive=True)
+        self.assertEqual(table["a"]["a"], "foo")
+        self.assertEqual(table["b"]["b"], "bar")
+        self.lua.globals()["data"] = table
+        self.lua.eval("assert(data.a.a=='foo', 'failed')")
```
scoder, line 603: Is there a test where we make sure that this raises an exception if the condition fails, so that the test actually fails?

lupa/tests/test.py
```diff
@@ -664,7 +721,8 @@ def test_setitem_array_none_initial(self):
         self.assertEqual(3, len(table))
 
     def test_setattr_table(self):
-        table = self.lua.eval('{ const={ name="Pi", value=3.1415927 }, const2={ name="light speed", value=3e8 }, val=1 }')
+        table = self.lua.eval(
+            '{ const={ name="Pi", value=3.1415927 }, const2={ name="light speed", value=3e8 }, val=1 }')
```
scoder, line 725: This doesn't seem necessary.

lupa/tests/test.py
```diff
@@ -2939,21 +3054,21 @@ def __call__(self, *args):
...
-        self.testmissingref({}, lambda o: str(o))                            # __tostring
-        self.testmissingref({}, lambda o: o[1])                              # __index
-        self.testmissingref({}, lambda o: lupa.as_itemgetter(o)[1])          # __index (itemgetter)
-        self.testmissingref({}, lambda o: lupa.as_attrgetter(o).items)       # __index (attrgetter)
-        self.testmissingref({}, lambda o: assign(o[1]))                      # __newindex
+        self.testmissingref({}, lambda o: str(o))  # __tostring
+        self.testmissingref({}, lambda o: o[1])  # __index
+        self.testmissingref({}, lambda o: lupa.as_itemgetter(o)[1])  # __index (itemgetter)
+        self.testmissingref({}, lambda o: lupa.as_attrgetter(o).items)  # __index (attrgetter)
+        self.testmissingref({}, lambda o: assign(o[1]))  # __newindex
```
scoder, line 3061: This kind of change seems unhelpful.

setup.py
```diff
@@ -333,7 +333,12 @@ def has_option(name):
     )]
 
 if cythonize is not None:
-    ext_modules = cythonize(ext_modules)
+    ext_modules = cythonize(ext_modules, compiler_directives={
+        "always_allow_keywords": True,
+        "cdivision": True,
+        "boundscheck": False,
+        "wraparound": False
+    })
```
scoder, line 341: This seems unrelated to this PR and is worth a discussion.

lupa/_lupa.pyx
```diff
@@ -1135,7 +1136,7 @@ cdef object resume_lua_thread(_LuaThread thread, tuple args):
             # already terminated
             raise StopIteration
         if args:
-            nargs = len(args)
+            nargs = <int>len(args)
```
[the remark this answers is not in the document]
synodriver: The msvc compiler just complains about the type convertion without this.

lupa/_lupa.pyx
```diff
@@ -1462,6 +1463,19 @@ cdef int py_to_lua(LuaRuntime runtime, lua_State *L, object o, bint wrap_none=Fa
...
         pushed_values_count = 1
+    elif recursive and isinstance(o, Sequence):
+        lua.lua_createtable(L, <int>len(o), 0)   # create a table at the top of stack, with narr already known
+        for i, v in enumerate(o):
+            py_to_lua(runtime, L, v, wrap_none, recursive)
+            lua.lua_rawseti(L, -2, i+1)
+        pushed_values_count = 1
+    elif recursive and isinstance(o, Mapping):
+        lua.lua_createtable(L, 0, <int>len(o))  # create a table at the top of stack, with nrec already known
+        for key, value in o.iteritems(): # to compatible with py2
```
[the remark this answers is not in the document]
synodriver: > 

I  saw the code below use o.iteritems...

lupa/tests/test.py
synodriver: > All whitespace changes are unrelated to this PR. It's mostly ok to make them (although it makes the git history more difficult to use), but I'd rather have them in a separate PR.

This is caused by the IDE... Would revert this later.

lupa/tests/test.py
synodriver: Ok, I would add it later

setup.py
synodriver: It just speed up generated code without extra check.

lupa/_lupa.pyx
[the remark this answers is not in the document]
synodriver: Would this ok?
```
cdef int _len_as_int(Py_ssize_t obj) except -1:
    if obj > <Py_ssize_t>LONG_MAX:
        raise OverflowError
    return <int>obj
```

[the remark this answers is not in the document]
synodriver: Maybe we should add ```inline``` to it.

synodriver:
Conflicts are resolved now.

lupa/_lupa.pyx
```diff
@@ -1462,13 +1467,26 @@ cdef int py_to_lua(LuaRuntime runtime, lua_State *L, object o, bint wrap_none=Fa
     elif isinstance(o, float):
         lua.lua_pushnumber(L, <lua.lua_Number><double>o)
         pushed_values_count = 1
+    elif isinstance(o, _PyProtocolWrapper):
+        type_flags = (<_PyProtocolWrapper> o)._type_flags
+        o = (<_PyProtocolWrapper> o)._obj
+        pushed_values_count = py_to_lua_custom(runtime, L, o, type_flags)
+    elif recursive and isinstance(o, Sequence):
```
scoder, line 1474: This just speeds up the type test for the "it'll usually be a list" case.
```suggestion
    elif recursive and isinstance(o, (list, Sequence)):
```

lupa/_lupa.pyx
```diff
@@ -1462,13 +1467,26 @@ cdef int py_to_lua(LuaRuntime runtime, lua_State *L, object o, bint wrap_none=Fa
...
+        lua.lua_createtable(L, _len_as_int(len(o)), 0)   # create a table at the top of stack, with narr already known
+        for i, v in enumerate(o):
+            py_to_lua(runtime, L, v, wrap_none, recursive)
+            lua.lua_rawseti(L, -2, i+1)
+        pushed_values_count = 1
+    elif recursive and isinstance(o, Mapping):
```
scoder, line 1480:
```suggestion
    elif recursive and isinstance(o, (dict, Mapping)):
```

lupa/_lupa.pyx
```diff
@@ -156,6 +157,10 @@ def lua_type(obj):
         lua.lua_settop(L, old_top)
         unlock_runtime(lua_object._runtime)
 
+cdef inline int _len_as_int(Py_ssize_t obj) except -1:
+    if obj > <Py_ssize_t>LONG_MAX:
+        raise OverflowError
+    return <int>obj
```
scoder, line 163: You could probably still get a C compiler warning here if `sizeof(long) == sizeof(Py_ssize_t)`, because it would detect that the condition is always false and, thus, useless.

BTW, why do you compare to `LONG_MAX` and not `INT_MAX` if you want to return an `int`?

lupa/_lupa.pyx
```diff
@@ -1462,13 +1467,26 @@ cdef int py_to_lua(LuaRuntime runtime, lua_State *L, object o, bint wrap_none=Fa
     elif isinstance(o, float):
         lua.lua_pushnumber(L, <lua.lua_Number><double>o)
         pushed_values_count = 1
+    elif isinstance(o, _PyProtocolWrapper):
+        type_flags = (<_PyProtocolWrapper> o)._type_flags
+        o = (<_PyProtocolWrapper> o)._obj
+        pushed_values_count = py_to_lua_custom(runtime, L, o, type_flags)
+    elif recursive and isinstance(o, Sequence):
+        lua.lua_createtable(L, _len_as_int(len(o)), 0)   # create a table at the top of stack, with narr already known
```
scoder, line 1475: Here's probably an error to handle if the allocation fails.

lupa/_lupa.pyx
```diff
@@ -1462,13 +1467,26 @@ cdef int py_to_lua(LuaRuntime runtime, lua_State *L, object o, bint wrap_none=Fa
...
+        for i, v in enumerate(o):
+            py_to_lua(runtime, L, v, wrap_none, recursive)
+            lua.lua_rawseti(L, -2, i+1)
```
scoder, line 1478:
```suggestion
        for i, v in enumerate(o, 1):
            py_to_lua(runtime, L, v, wrap_none, recursive)
            lua.lua_rawseti(L, -2, i)
```

lupa/_lupa.pyx
synodriver: Is there a better way to deal with it or we should direct cast it?

lupa/_lupa.pyx
synodriver: Is that used to prevent from over nesting？

synodriver: ```lua_createtable``` has no return value, so how do we know if it fails?

scoder:
Rather than requiring users to think about a hard depth limit for their data structures, what do you think about keeping a dict of mapped objects (by original object ID) and reusing already mapped object references? That way, we could really map recursive structures 1:1 without the risk to run into infinite loops while mapping.

synodriver:
> Rather than requiring users to think about a hard depth limit for their data structures, what do you think about keeping a dict of mapped objects (by original object ID) and reusing already mapped object references? That way, we could really map recursive structures 1:1 without the risk to run into infinite loops while mapping.

Great idea, I've implemented it in the latest commit.

lupa/_lupa.pyx
```diff
@@ -520,15 +525,15 @@ cdef class LuaRuntime:
...
+    def table_from(self, *args, bint recursive=False):
         """Create a new table from Python mapping or iterable.
 
         table_from() accepts either a dict/mapping or an iterable with items.
         Items from dicts are set as key-value pairs; items from iterables
         are placed in the table in order.
 
         Nested mappings / iterables are passed to Lua as userdata
-        (wrapped Python objects); they are not converted to Lua tables.
+        (wrapped Python objects) if `recursive` is False, they are not converted to Lua tables.
```
scoder, line 536: I think this is what you mean, right?
```suggestion
        (wrapped Python objects).  If `recursive` is False, they are not converted to Lua tables.
```

lupa/_lupa.pyx
```diff
@@ -1597,13 +1602,40 @@ cdef int py_to_lua(LuaRuntime runtime, lua_State *L, object o, bint wrap_none=Fa
...
+            lua.lua_createtable(L, 0, _len_as_int(len(o)))  # create a table at the top of stack, with nrec already known
+            mapped_objs[id(o)] = lua.lua_gettop(L)
+            for key, value in o.items():
+                py_to_lua(runtime, L, key, wrap_none, recursive, mapped_objs)
+                py_to_lua(runtime, L, value, wrap_none, recursive, mapped_objs)
+                lua.lua_rawset(L, -3)
+        else: # self-reference detected
+            idx = mapped_objs[id(o)]
+            lua.lua_pushvalue(L, idx)
+        pushed_values_count = 1
```
scoder, line 1635: I'd reverse the conditions to put the simple "already mapped" case first.
```suggestion
    elif recursive and isinstance(o, (list, Sequence)):
        if mapped_objs is None:
            mapped_objs = {}
        if id(o) in mapped_objs:
            obj = mapped_objs[id(o)]
            lua.lua_pushvalue(L, obj)
        else:
            lua.lua_createtable(L, _len_as_int(len(o)), 0)  # create a table at the top of stack, with narr already known
            mapped_objs[id(o)] = lua.lua_gettop(L)
            for i, v in enumerate(o, 1):
                py_to_lua(runtime, L, v, wrap_none, recursive, mapped_objs)
                lua.lua_rawseti(L, -2, i)
        pushed_values_count = 1
    elif recursive and isinstance(o, (dict, Mapping)):
        if mapped_objs is None:
            mapped_objs = {}
        if id(o) in mapped_objs:
            obj = mapped_objs[id(o)]
            lua.lua_pushvalue(L, obj)
            lua.lua_createtable(L, 0, _len_as_int(len(o)))  # create a table at the top of stack, with nrec already known
            mapped_objs[id(o)] = lua.lua_gettop(L)
            for key, value in o.items():
                py_to_lua(runtime, L, key, wrap_none, recursive, mapped_objs)
                py_to_lua(runtime, L, value, wrap_none, recursive, mapped_objs)
                lua.lua_rawset(L, -3)
        pushed_values_count = 1
```
Actually, I'd even reuse the existing `LuaRuntime.table_from()` implementation here. Or, rather, extract a `cdef` function from it and use it in both places. This seems needlessly duplicated code.

lupa/__init__.py
```diff
@@ -63,8 +63,10 @@ def _import_newest_lib():
         raise RuntimeError("Failed to import Lupa binary module.")
     # prefer Lua over LuaJIT and high versions over low versions.
     module_name = max(modules, key=lambda m: (m[1] == 'lua', tuple(map(int, m[2] or '0'))))
-
-    _newest_lib = __import__(module_name[0], level=1, fromlist="*", globals=globals())
+    try:
+        _newest_lib = __import__(module_name[0], level=1, fromlist="*", globals=globals())
+    except ModuleNotFoundError:
+        _newest_lib = __import__(module_name[1], level=1, fromlist="*", globals=globals())
```
scoder, line 69: This seems unrelated. Why did you change it?

scoder:
I've extracted the logic from `LuaRuntime.table_from()` in the master branch to make it reusable for the recursive container mappings. Could you please adapt your code? Most of the changes in `py_to_lua` can probably be avoided now.

lupa/__init__.py
synodriver: Actually this is related to another bug. If you link to your custom lua libs, the original will fail to find corret module.

synodriver:
> I've extracted the logic from `LuaRuntime.table_from()` in the master branch to make it reusable for the recursive container mappings. Could you please adapt your code? Most of the changes in `py_to_lua` can probably be avoided now.

I'm afraid not because ```py_to_lua_table``` still make calls to ```py_to_lua```

scoder:
> > Most of the changes in `py_to_lua` can probably be avoided now.
> 
> I'm afraid not because `py_to_lua_table` still make calls to `py_to_lua`

Sorry, I just meant the additions, not the flags etc. Both functions need to call each other, but you can avoid the dict/list mappings in `py_to_lua` now.

synodriver:
> avoid the dict/list mappings in `py_to_lua` now.

Could you please give me some advice about that? Perhaps I misunderstood your idea, but I think there might be dict in dict, so we have to check them in ```py_to_lua``` again.

scoder:
> > avoid the dict/list mappings in `py_to_lua` now.
> 
> Could you please give me some advice about that? Perhaps I misunderstood your idea, but I think there might be dict in dict, so we have to check them in `py_to_lua` again.

I think it mostly boils down to something like this:
```python
     elif recursive and isinstance(o, (list, dict, Sequence, Mapping)):
        # check existing mapped_objs cache, then failing that:
        table = py_to_lua_table([o], recursive=True)
        # update cache
```

synodriver:
Finally... I made it, looks better now.

lupa/tests/test.py
```diff
@@ -632,6 +632,91 @@ def test_table_from_table_iter_indirect(self):
...
+                if type(v) == "table" then
+                    itertable(v)
+                else
+                    print(v)
+                end  
+            end 
+        end
+        print('\\n')
+        itertable(data)
+        """)
```
scoder, line 657: The output of these is only printed into the test log and not validated anywhere. Maybe you could collect the output in a list and return it to Python for validation, instead of printing it?

lupa/tests/test.py
```diff
@@ -632,6 +632,91 @@ def test_table_from_table_iter_indirect(self):
...
+        del self.lua.globals()["data"]
```
scoder, line 658: This isn't necessary. Each test will get a freshly created LuaRuntime that gets cleaned up automatically after the test run.

scoder (commented):
Thanks, that removes a lot of duplicated functionality. Looks much better now.

lupa/_lupa.pyx
```diff
@@ -1667,13 +1678,20 @@ cdef _LuaTable py_to_lua_table(LuaRuntime runtime, lua_State* L, items):
...
     lua.lua_newtable(L)
     # FIXME: how to check for failure?
-
+    if recursive and mapped_objs is None:
+        mapped_objs = {}
     try:
         for obj in items:
+            if recursive:
+                if id(obj) not in mapped_objs:
+                    mapped_objs[id(obj)] = lua.lua_gettop(L)
```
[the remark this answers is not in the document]
synodriver: Because the call is recursive. If the cache is moved to the end of the loop body, ```py_to_lua``` might trigger a call to  
 ```py_to_lua_table``` again during the loop, which triggers another call and so on, so it might never be able to called.

lupa/tests/test.py
synodriver: It was used for debugging. I'll fix it.

lupa/_lupa.pyx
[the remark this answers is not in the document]
synodriver: And, this function creates a table at the beginning, which will be filled with data from python objects, and that table will be converted into a ```_LuaTable```, so we just need to keep an eye on that table.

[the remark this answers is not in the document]
scoder: Could you try mapping a Python list like `[[3, 3, 3]]` to Lua with this? I.e. one that actually repeats the same value multiple times, so that the cache would be hit? It looks like it might end up referencing the mapped table itself as a value in the Lua table at some point.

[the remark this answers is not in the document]
synodriver: ```[[3, 3, 3]]``` have a depth of 2, it will call ```py_to_lua_table``` twice. Each time the id of the list is cached, and at the second time, after the inner list is cached, it just goes into the list handler of the loop, without caching the PyLongObject. That means, only list and dict is cached.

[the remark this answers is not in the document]
synodriver: A new test have been added. It works as expected.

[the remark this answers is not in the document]
scoder: > And, this function creates a table at the beginning, which will be filled with data from python objects, and that table will be converted into a `_LuaTable`, so we just need to keep an eye on that table.

Ok, so, do I understand correctly that the invariant in this for-loop is that the top of the Lua stack always refers to the Lua table that we are filling? And that is the value that we put into `mapped_objs`?

Why don't you read the stack top once, before the loop, give it a name (like `lua_table_ref`) and use that instead? That would be much clearer than reading seemingly randomly from the top of the Lua stack and putting whatever we get into `mapped_objs` for an object ID that we have never seen before.

lupa/_lupa.pyx
```diff
@@ -1667,13 +1678,20 @@ cdef _LuaTable py_to_lua_table(LuaRuntime runtime, lua_State* L, items):
...
+                else:
+                    idx = mapped_objs[id(obj)]
+                    return new_lua_table(runtime, L, <int>idx)
             if isinstance(obj, dict):
                 for key, value in (<dict>obj).items():
-                    py_to_lua(runtime, L, key, wrap_none=True)
-                    py_to_lua(runtime, L, value)
+                    py_to_lua(runtime, L, key, True, recursive, mapped_objs)
+                    py_to_lua(runtime, L, value, False, recursive, mapped_objs)
```
scoder, line 1694: As I suggested before, please use keyword arguments instead of bare `True`/`False` values.

lupa/tests/test.py
```diff
@@ -598,10 +601,16 @@ def test_table_from_bad(self):
         self.assertRaises(TypeError, self.lua.table_from, None)
         self.assertRaises(TypeError, self.lua.table_from, {"a": 5}, 123)
 
-    # def test_table_from_nested(self):
-    #     table = self.lua.table_from({"obj": {"foo": "bar"}})
-    #     lua_type = self.lua.eval("type")
-    #     self.assertEqual(lua_type(table["obj"]), "table")
+    def test_table_from_nested(self):
+        table = self.lua.table_from([[3, 3, 3]], recursive=True)
```
scoder, line 605: This may not have been the best example, just something that I came up with trying to understand your implementation. However, a more deeply nested example would be good, to make sure that, say, a dict of lists of dicts of dicts of lists of lists also works, or a (list) table of (mapping) tables. You're only testing one level of nesting, always starting with a dict. Maybe you can generate a data structure (or a few), map it back and forth, and then compare the result to itself? That would allow testing something larger and deeper.

lupa/__init__.py
synodriver: Ok, I will put it in a seperate pr.

lupa/tests/test.py
synodriver: Accutally there are other tests which are far more deeper, for instance, the ```test_table_from_self_ref_obj```, which is a infinite deep structure.

lupa/_lupa.pyx
synodriver: Ok, I've changed them all.

lupa/_lupa.pyx
```diff
@@ -1666,14 +1678,24 @@ cdef _LuaTable py_to_lua_table(LuaRuntime runtime, lua_State* L, items):
     check_lua_stack(L, 5)
     old_top = lua.lua_gettop(L)
     lua.lua_newtable(L)
+    cdef int lua_table_ref = lua.lua_gettop(L) # the index of the lua table which we are filling
     # FIXME: how to check for failure?
```
scoder, line 1682: I think this comment was referring to the line above. And it's probably outdated, because we can't check for failures, Lua would just jump to the error handler.
```suggestion
    # FIXME: handle allocation errors
    cdef int lua_table_ref = lua.lua_gettop(L) # the index of the lua table which we are filling
```

lupa/_lupa.pyx
```diff
@@ -1666,14 +1678,24 @@ cdef _LuaTable py_to_lua_table(LuaRuntime runtime, lua_State* L, items):
...
     try:
         for obj in items:
+            if recursive:
+                if id(obj) not in mapped_objs:
+                    # this object is never seen before, we should cache it
+                    mapped_objs[id(obj)] = lua_table_ref
+                else:
+                    # this object has been cached, just get the corresponding lua table's index
+                    idx = mapped_objs[id(obj)]
+                    return new_lua_table(runtime, L, <int>idx)
```
scoder, line 1694: Ah, thanks for that comment, that actually helps. Then you're not using `mapped_objs` as it's named but rather as a `mapped_tables`, right? Seems worth renaming then to make the code clearer.

lupa/tests/test.py
synodriver: Ok, I'll try to cover more tests, like ```List[Dict]```, Dict[str, Dict]```, List[List]```, Dict[str, List]```

lupa/_lupa.pyx
```diff
@@ -520,21 +525,22 @@ cdef class LuaRuntime:
...
         """Create a new table from Python mapping or iterable.
 
         table_from() accepts either a dict/mapping or an iterable with items.
         Items from dicts are set as key-value pairs; items from iterables
         are placed in the table in order.
 
         Nested mappings / iterables are passed to Lua as userdata
-        (wrapped Python objects); they are not converted to Lua tables.
+        (wrapped Python objects). If `recursive` is False (the default),
+        they are not converted to Lua tables.
```
scoder, line 537: Actually, the whole paragraph is not clear due to a perceived double negation.
```suggestion
        (wrapped Python objects) by default.  If `recursive` is True,
        they are converted to Lua tables recursively, handling loops
        and duplicates via de-duplication.
```

lupa/tests/test.py
scoder: What about something like this:
```python
def test():
    from itertools import count
    def make_ds(*children):
        for child in children:
            yield child
        yield list(children)
        yield dict(zip(count(), children))
        yield {chr(ord('A') + i): child for i, child in enumerate(children)}

    for ds1 in make_ds(1, 2, 'x', 'y'):
        for ds2 in make_ds(ds1):
            for ds in make_ds(ds1, ds2):
                table = self.lua.table_from(ds)
                # validate table == ds
```
I hope it's mostly clear, the idea is to generate 0-2 level data structures of all sorts of nesting combinations, and then validate that the result looks the same in Lua as in Python. The validation would probably involve a couple of type checks and back-mappings Lua->Python, but shouldn't be too complex. The above loops generate somewhat redundant results, AFAICT, but not too bad. Maybe you can come up with something smarter even.

synodriver: Opps... I found a bug, and it can be reproduced in master branch, just use ```lua.table_from(1)```, in this way ```py_to_lua_table``` got a ```(1, )```, and as it iterate over the tuple, the ```PyLongObject``` is passed to 
```
 else:
       for arg in obj:
```
but ```int``` is not iterable. Should fix it somehow so that I can continus with my pr.

synodriver: The structures generated by this function should also be applied to master branch I think.

synodriver: By the way, there might be some gc problem with cpython itself, in the loop test there are some lua tables can't match the corresponding pyobject, but when I test them in a individual test, the problem just gone.

synodriver:
Iet me explaint what I found these days. There is a bug in master branch which can not pass your testcase.
```
lua.table_from(1)
```
will always end up with an error: ```int``` is not iterable.
At the same time, there might be some gc problem with cpython itself, in the loop test there are some lua tables can't match the corresponding pyobject, but when I test them in a individual test, the problem just disappear, making it very hard to debug.

scoder:
> There is a bug in master branch which can not pass your testcase.
> ```
> lua.table_from(1)
> ```

That's not a bug. The method expects one or more iterables or mappings as arguments, not plain values. I.e., you can write `lua.from_table([1], [2], [3,4,5])`, but not `lua.from_table(1,2,3,4,5)`. That's also what the docstring says, at least as I read it.

lupa/tests/test.py
scoder: My test code proposal was buggy. Here's a revised version that does not generate single elements:
```python
def test():
    from itertools import count, chain
    def make_ds(*children):
        yield list(children)
        yield dict(zip(count(), children))
        yield {chr(ord('A') + i): child for i, child in enumerate(children)}

    elements = [1, 2, 'x', 'y']
    for ds1 in make_ds(*elements):
        for ds2 in make_ds(ds1):
            for ds3 in make_ds(ds1, elements, ds2):
                for ds in make_ds(ds1, ds2, ds3):
                    table = self.lua.table_from(ds)
                    # validate table == ds
```

scoder: I added this test to the master branch and merged it into this PR. Since the "validation" in the master branch is really rudimentary, it hopefully won't fail here, but can now be adapted to validate the transitive translation.

synodriver:
Thanks for your help with the nested data structure generator, it looks better now. The last thing to do is figure out a way to compare them without encountering strange garbage collection bugs in CPython.....

scoder:
Thanks for all the work on this. I think it's good enough to be released in Lupa 2.1.

PR #212, opened 2022-05-08
Add option to limit the memory usage of Lua code

### LINKED ISSUE #211: Option to limit memory usage
I'd love to see an option to limit the max memory usage of a single LuaRuntime.

Lua supports passing a custom allocator function to [`lua_newstate`](https://www.lua.org/manual/5.1/manual.html#lua_newstate) which could refuse to allocate more memory when the limit is reached.

An example of this can be found [here on stackoverflow](https://stackoverflow.com/questions/9671793/limiting-a-lua-scripts-memory-usage).

Le0Developer:
Closes #211.


lupa/_lupa.pyx
```diff
@@ -1609,6 +1642,40 @@ cdef call_lua(LuaRuntime runtime, lua_State *L, tuple args):
...
+
+    if ptr is NULL:
+        # <http://www.lua.org/manual/5.2/manual.html#lua_Alloc>:
+        # When ptr is NULL, osize encodes the kind of object that Lua is
+        # allocating.
+        # Since we don’t care about that, just mark it as 0.
+        osize = 0
+
+    if nsize == 0:
+        free(ptr)
```
scoder, line 1657: We shouldn't use plain C memory allocation. If we do the allocation ourselves, we should either do what Lua does internally, and if that really is plain C allocation, then we should use Python's memory allocator instead.

lupa/_lupa.pyx
```diff
@@ -1609,6 +1642,40 @@ cdef call_lua(LuaRuntime runtime, lua_State *L, tuple args):
...
+        return ptr
+
+cdef int _lua_panic(lua_State *L) nogil:
+    cdef char* msg = lua.lua_tostring(L, -1)
+    if msg == NULL:
+        msg = "error object is not a string"
+    cdef char* message = "PANIC: unprotected error in call to Lua API (%s)\n"
+    fprintf(stderr, message, msg)
+    fflush(stderr)
+    return 0 # return to Lua to abort
```
scoder, line 1677: I admit that jumping to address `1` (as done before) isn't exactly a great error handler, but it feels like we should generally improve the way fatal errors are handled, rather than making them fatal but just in other ways.

OTOH, this is an improvement all by itself, so… shouldn't it be a separate PR?

lupa/_lupa.pyx
```diff
@@ -1609,6 +1642,40 @@ cdef call_lua(LuaRuntime runtime, lua_State *L, tuple args):
...
+        return NULL
+    elif nsize == osize:
+        return ptr
+    else:
+        if nsize > osize and left[0] < nsize - osize: # reached the limit
+            return NULL
+        ptr = realloc(ptr, nsize)
+        if ptr: # reallocation successful?
+            left[0] -= nsize + osize
+        return ptr
```
scoder, line 1668: We leak the old pointer here if the allocation fails.

lupa/_lupa.pyx
```diff
@@ -460,6 +486,13 @@ cdef class LuaRuntime:
             lua.lua_settop(L, old_top)
             unlock_runtime(self)
 
+    def set_max_memory(self, max_memory):
```
scoder, line 489: Not sure if we need this. Do you see a use case for changing the memory limit later on?

lupa/_lupa.pyx
```diff
@@ -242,12 +248,24 @@ cdef class LuaRuntime:
     cdef object _attribute_getter
     cdef object _attribute_setter
     cdef bint _unpack_returned_tuples
+    cdef size_t _max_memory
+    cdef size_t* _memory_left
```
scoder, line 252: Why not use a plain `size_t` integer attribute here and take the address of that, instead of allocating memory and storing the pointer to that?

lupa/_lupa.pyx
Le0Developer: This error handler is the one that luaL_newstate registers

lupa/_lupa.pyx
Le0Developer: I'm using this in the tests atm

lupa/_lupa.pyx
Le0Developer: > if that really is plain C allocation

It is. This is the function `luaL_newstate` uses:
```c
     static void *l_alloc (void *ud, void *ptr, size_t osize,
                                                size_t nsize) {
       (void)ud;  (void)osize;  /* not used */
       if (nsize == 0) {
         free(ptr);
         return NULL;
       }
       else
         return realloc(ptr, nsize);
     }
```

Le0Developer: Python's memory allocator relies on GIL

Le0Developer: When using python's allocator, I'm getting sporadic segfaults and this: `malloc: *** error for object 0x7fec8551b280: pointer being freed was not allocated`

lupa/_lupa.pyx
Le0Developer: Using this to temporarily remove the limit to load trusted library and code and then restricting it once loading a user's code

lupa/_lupa.pyx
```diff
@@ -242,14 +253,22 @@ cdef class LuaRuntime:
...
                   bint register_eval=True, bint unpack_returned_tuples=False,
-                  bint register_builtins=True, overflow_handler=None):
-        cdef lua_State* L = lua.luaL_newstate()
+                  bint register_builtins=True, overflow_handler=None,
+                  max_memory=0):
+        cdef lua_State* L
+
+        self._memory_left = 0
+
+        L = lua.lua_newstate(<lua.lua_Alloc>&_lua_alloc_restricted, &self._memory_left)
```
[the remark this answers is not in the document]
Le0Developer: > Do I understand correctly that the allocator function is not needed when no memory limit is used? Then we could keep the old call in the normal (no limit) case.
> I'd really like to get rid of the set_max_memory() method, if possible. It feels better to make this an init time only setting, because the effect of changing the maximum is really indirect. As your test shows, it only applies when the next allocation happens (which might be at unexpected times) and does not fail immediately, even though we might already have allocated more memory than the new limit.

I have to use the allocator function because `set_max_memory()` might be called in the future.

In my use case, I have to load multiple libraries just like the stdlib before running the actual code which should not count towards the memory limit of the code.

Maybe, the default could be None in which case the default `luaL_newstate` allocator will be used and `set_max_memory()` will throw? ([like how it was previously](https://github.com/scoder/lupa/pull/212/commits/2669b6346b574d502cfbbcf2b0f79d3d3bf32d0b#diff-cf9743b035eef33132b4cdec8aad44f5b0c2dfa113c858624a0bd3853e2b6604R254))

> BTW, why is it runtime.max_memory for reading but runtime.set_max_memory() for writing? Why not use a property for both?

Personal preference, and a relic from earlier testing stuff. (max_memory allocator was optional earlier and this function [threw an exception](https://github.com/scoder/lupa/pull/212/commits/bbfaf16b14d0d25a7dcb28d9eb8c50b81fdd3756#diff-cf9743b035eef33132b4cdec8aad44f5b0c2dfa113c858624a0bd3853e2b6604R486) and setting properties really shouldn't throw an exception imo)

Le0Developer:
It seems like LuaJIT does not support a custom allocator in `lua_newstate` on 64bit.

See the following and failing tests:

> Make sure you use luaL_newstate. Avoid using lua_newstate, since this uses the (slower) default memory allocator from your system (no support for this on 64 bit architectures).
- https://luajit.org/install.html

lupa/_lupa.pyx
scoder: Ok, then ISTM that we need to handle whatever Lua returns us here, which are `malloc()` allocated memory blocks.
(And yes, Python's allocator requires the GIL and does more than just calling `malloc` and friends, potentially much more.)

scoder:
> It seems like LuaJIT does not support a custom allocator in `lua_newstate` on 64bit.

Hmm. That is annoying. Then we should raise an exception when we detect that case and users request a custom allocator.

Le0Developer:
Setting max_memory to None will now use the default lua allocator. Also added documentation for `set_max_memory()` and `max_memory`.

`set_max_memory()` also a strict option now, which will raise an exception if you try to set max_memory to a value lower than currently used.

Le0Developer:
Just tested running LuaJIT 2.1.0-beta3 and it works on 64bit. 🤷 

Following is a patch that disallows LuaJIT on 64bit, if needed:

<details>

```patch
diff --git a/lupa/lua.pxd b/lupa/lua.pxd
index 35d4b6e15a4fc18e3deb20353be72cae4ebaee86..72cdce030bfaf13db38af6d97aed66c85f55eacf 100644
--- a/lupa/lua.pxd
+++ b/lupa/lua.pxd
@@ -266,6 +266,13 @@ cdef extern from "lua.h" nogil:
         int i_ci               #           active function */
 
 
+# luajit.h
+
+cdef extern from "luajit.h" nogil:
+    int LUAJIT_VERSION_NUM
+
 ################################################################################
 # lauxlib.h
 ################################################################################
@@ -465,11 +472,18 @@ cdef extern from * nogil:
     #if LUA_VERSION_NUM < 502
     #define lua_pushglobaltable(L)  lua_pushvalue(L, LUA_GLOBALSINDEX)
     #endif
+
+    #if defined LUAJIT_VERSION_NUM && UINTPTR_MAX == 0xffffffffffffffff
+    #define IS_LUAJIT64 1
+    #else
+    #define IS_LUAJIT64 0
+    #endif
     """
     int read_lua_version(lua_State *L)
     int lua_isinteger(lua_State *L, int idx)
     lua_Integer lua_tointegerx (lua_State *L, int idx, int *isnum)
     void lua_pushglobaltable (lua_State *L)
+    char IS_LUAJIT64
 
 cdef extern from *:
     # Limits for Lua integers (in Lua<5.3: PTRDIFF_MIN, PTRDIFF_MAX)
diff --git a/lupa/_lupa.pyx b/lupa/_lupa.pyx
index 22656680343b1c6e9f97ef8adf972df9958e614a..c173d9a9376339e2dfec982e04ca5ca9b7a8b6d2 100644
--- a/lupa/_lupa.pyx
+++ b/lupa/_lupa.pyx
@@ -263,6 +263,9 @@ cdef class LuaRuntime:
                   bint register_eval=True, bint unpack_returned_tuples=False,
                   bint register_builtins=True, overflow_handler=None,
                   max_memory=None):
+        if lua.IS_LUAJIT64 and max_memory is not None:
+            raise RuntimeError("max_memory is not supported on 64bit LuaJIT")
+        
         cdef lua_State* L
 
         self._memory_left = 0

```

</details>

scoder:
>`set_max_memory()` also a strict option now, which will raise an
>exception if you try to set max_memory to a value lower than currently used.

I'd rather expose the current memory usage as readonly attribute and let users do their own checks. That seems more generally useful.

Le0Developer:
Added `memory_used`

scoder:
There's still a test failure. Maybe you could look into that? I'd like to release Lupa 2.0 soon and this would be a nice addition.

Le0Developer:
Unable to reproduce the test failure locally but I think that should've fixed it.

lupa/_lupa.pyx
```diff
@@ -1631,8 +1712,10 @@ cdef run_lua(LuaRuntime runtime, bytes lua_code, tuple args):
     try:
         check_lua_stack(L, 1)
         if lua.luaL_loadbuffer(L, lua_code, len(lua_code), '<python>'):
-            raise LuaSyntaxError(build_lua_error_message(
-                runtime, L, err_message=u"error loading code: %s"))
+            error = build_lua_error_message(runtime, L)
+            if error == "not enough memory":
+                raise LuaMemoryError(error)
```
scoder, line 1717: It seems that "not enough memory" isn't always the (complete) error message here. We need either a safer or more relaxed way to detect memory errors.
```
ERROR: test_not_enough_memory (lupa.tests.TestMaxMemory_lua53)
----------------------------------------------------------------------
  File "/home/runner/work/lupa/lupa/lupa/tests/test.py", line 3019, in test_not_enough_memory
    self.assertRaises(lupa.LuaMemoryError, self.lua.eval, "('a'):rep(50000)")
  File "/opt/hostedtoolcache/Python/2.7.18/x64/lib/python2.7/unittest/case.py", line 473, in assertRaises
    callableObj(*args, **kwargs)
  File "lupa/lua53.pyx", line 395, in lupa.lua53.LuaRuntime.eval
    return run_lua(self, b'return ' + lua_code, args)
  File "lupa/lua53.pyx", line 1721, in lupa.lua53.run_lua
    return call_lua(runtime, L, args)
  File "lupa/lua53.pyx", line 1735, in lupa.lua53.call_lua
    return execute_lua_call(runtime, L, len(args))
  File "lupa/lua53.pyx", line 1798, in lupa.lua53.execute_lua_call
    raise_lua_error(runtime, L, result_status)
  File "lupa/lua53.pyx", line 1650, in lupa.lua53.raise_lua_error
    raise LuaError(build_lua_error_message(runtime, L))
LuaError: [string "<python>"]:1: not enough memory for buffer allocation
stack traceback:
	[string "<python>"]:1: in main chunk
	[C]: in function 'string.rep'
```

scoder (commented):
There are still test failures. Most of them are related to LuaJIT, and since you noted that LuaJIT is not supported here, we need to skip your tests when running the tests against LuaJIT. A test method decorator like this might work:
```python
def no_luajit(test_method):
    @functools.wraps(test_method)
    def test(self):
        if "luajit" in self.lupa.LuaRuntime().lua_implementation().lower():
            self.skipTest("not supported in LuaJIT")
        return test_method(self)
    return test
```

lupa/_lupa.pyx
Le0Developer:
```suggestion
            if error.startswith("not enough memory"):
                raise LuaMemoryError(error)
```

scoder: I think the error message above starts with this: "[string "<python>"]:1: not enough memory for buffer allocation"

scoder: (You may have to merge the master branch into yours, as the CI runner does.)

scoder: Isn't there an error code somewhere in Lua that we could look at instead? And that would allow us to safely detect a specific type of error?

scoder: Actually, we already do look at the error codes, in `raise_lua_error()`. When we see a `LUA_ERRMEM`, we raise a `MemoryError`.
https://github.com/scoder/lupa/blob/c92dc8208320ca0f0757e2feac5347b26cf92916/lupa/_lupa.pyx#L1557-L1563

scoder: And I just noticed that we previously raised a standard Python `MemoryError` for memory allocation failures. Switching to a `LuaMemoryError` that does not inherit from `MemoryError` is an incompatible change. I don't see a reason to do that. We can just continue to raise `MemoryError()` (without message), I think.

Le0Developer: Python's MemoryError does not inherit from Lupa's LuaError, so try/catchs with only LuaError may not catch it.
I'd prefer changing LuaMemoryError to inherit from MemoryError.

(also thanks for github for not loading comments, so I almost missed all your comments when I resolved this)

Le0Developer:
I am getting no more test failures locally (running with 53 and 54)

lupa/tests/test.py
```diff
@@ -3009,27 +3009,33 @@ def load_tests(loader, standard_tests, pattern):
 class TestMaxMemory(SetupLuaRuntimeMixin, LupaTestCase):
     lua_runtime_kwargs = {"max_memory": 10000}
 
+    def setUp(self):
+        # need to test in here because the creation of the LuaRuntime fails
+        if "luajit" in self.lupa.LuaRuntime().lua_implementation.lower():
+            self.skipTest("not supported in LuaJIT")
+        return super().setUp()
```
scoder, line 3016: `super()` requires arguments in Py2.7.

lupa/_lupa.pyx
```diff
@@ -419,7 +419,8 @@ cdef class LuaRuntime:
             else:
                 err = lua.lua_tolstring(L, -1, &size)
                 error = err[:size] if self._encoding is None else err[:size].decode(self._encoding)
-                if error == (b"not enough memory" if self._encoding is None else "not enough memory"):
+                not_enough_memory = (b"not enough memory" if self._encoding is None else "not enough memory")
+                if error.startswith(not_enough_memory):
```
scoder, line 423: This should probably also be `in`, no?

scoder: (probably worth adding a test for as well)

lupa/tests/test.py
Le0Developer: How long will lupa still support py2? It's been EOL for over 1.5 years.

lupa/tests/test.py
```diff
@@ -3030,7 +3030,7 @@ def test_decrease_memory(self):
         self.lua.set_max_memory(10000)
         self.assertGreater(self.lua.max_memory, 10000)
         self.assertGreaterEqual(self.lua.memory_used, 50000)
-        self.assertRaises(self.lupa.LuaMemoryError, self.lua.eval, "('a'):rep(10)")
+        self.assertRaises(self.lupa.LuaMemoryError, self.lua.eval, "('b'):rep(10)")
```
Le0Developer, line 3033: Honestly no idea why this keeps failing sometimes [on windows](https://github.com/scoder/lupa/runs/7715974364?check_suite_focus=true).

Might be string reuse (changed to `b` to check) or maybe it puts small values on stack and not heap which doesn't call our custom allocator?

scoder:
>How long will lupa still support py2? It's been EOL for over 1.5 years.

Which doesn't mean that people aren't using it any more. I'll support Py2 as long as it's easy to do, which it still is.

CHANGES.rst
```diff
@@ -11,6 +11,9 @@ Lupa change log
   Lua 5.4, LuaJIT 2.0 and LuaJIT 2.1 beta.  Note that this is build specific
   and may depend on the platform.  A normal Python import cascade can be used.
 
+* GH#211: Option to limit memory usage
```
Le0Developer, line 14: reflect pr title change here too?

Le0Developer:
```suggestion
* GH#211: Add option to limit the memory usage of Lua code
```

lupa/_lupa.pyx
```diff
@@ -1725,6 +1817,44 @@ cdef tuple unpack_multiple_lua_results(LuaRuntime runtime, lua_State *L, int nar
...
+        return NULL
+    elif new_size == old_size:
+        return ptr
+    else:
+        if memory_left[0] > 0 and new_size > old_size and memory_left[0] <= new_size - old_size:  # reached the limit
+            return NULL
+        new_ptr = realloc(ptr, new_size)
+        if new_ptr is not NULL and memory_left[0] > 0:
+            memory_left[0] += new_size - old_size
+        return new_ptr
```
scoder, line 1845: I changed two things here.

1) `realloc()` does not `free()` the memory on failures, and I think we shouldn't do that here either. According to the [example in the Lua docs](https://www.lua.org/manual/5.4/manual.html#lua_Alloc), Lua uses the plain `realloc()` behaviour. Therefore, I removed the `free()`. I would assume Lua to call into our allocator again in order to free the old pointer on failures, if it decides to free it.

2) The original calculation was `memory_left[0] -= new_size + old_size`, which looked wrong to me. It should add the difference, not subtract the sum.

Le0Developer: You are correct - the logic is wrong, but so is yours.
It should be `memory_left[0] -= new_size - old_size`.

```py
# allocate 20 more bytes
memory_left, new, old = 100, 50, 30
print(memory_left - (new - old))  # 80 (correct)
print(memory_left - (new + old))  # 20
print(memory_left + (new - old))  # 120
print(memory_left + (new + old))  # 180
```

scoder: Right. It says "memory left", not "memory used". I mixed that up.

lupa/_lupa.pyx
```diff
@@ -1821,28 +1821,32 @@ cdef tuple unpack_multiple_lua_results(LuaRuntime runtime, lua_State *L, int nar
...
 
+    cdef void* new_ptr
     if new_size == 0:
         free(ptr)
-        if memory_left[0] > 0:
-            memory_left[0] += old_size  # add old size to available memory
-        return NULL
+        if memory_is_counted:
+            memory_left += old_size  # add deallocated old size to available memory
+        new_ptr = NULL
```
Le0Developer, line 1839: This is never commited to `memory_left_status[0]` because `new_ptr` Is NULL and the assignment at the end of the function is skipped.
```suggestion
            memory_left_status[0] += old_size  # add deallocated old size to available memory
```

lupa/_lupa.pyx
scoder: While cleaning it up a bit, I noticed that we stop counting the memory when users disable it, but then, when users enable it again, we start from whatever "memory left" status we had when it was disabled. That doesn't sound clean. The current "memory left" might be arbitrarily out of sync with the actual memory usage at that point. Especially when you think of future memory deallocations of memory that was allocated during the time without memory limits.

Why not *always* count the memory that is *used* instead of (most of the time) the memory that is *left*, and then compare it to whatever the maximum allows memory usage is at the point of the allocation?

lupa/_lupa.pyx
```diff
@@ -1725,6 +1817,48 @@ cdef tuple unpack_multiple_lua_results(LuaRuntime runtime, lua_State *L, int nar
...
+    elif new_size == old_size:
+        new_ptr = ptr
+    else:
+        if memory_is_counted and new_size > old_size and memory_left <= new_size - old_size:  # reached the limit
+            return NULL
+        new_ptr = realloc(ptr, new_size)
+
+    if new_ptr is not NULL and memory_is_counted:
```
scoder, line 1847:
```suggestion
    if memory_is_counted:
```

lupa/_lupa.pyx
scoder: Hmm. Right. I was intending the final update to depend only on `memory_is_counted`, but that doesn't work in the case that `realloc()` fails. :-/

lupa/_lupa.pyx
scoder: Or, assume that the current memory usage is always `0` when users set a memory limit, and just take care of the difference from that point on.

Le0Developer: Because it was easier - `ud` can only point to one thing and I chose an size_t for that in the beginning when things were simpler.
We could probably change it to point to the LuaRuntime struct?

scoder: > We could probably change it to point to the LuaRuntime struct?

That won't work because the allocator needs to run outside of the GIL. But we could use our own "memory allocation status struct", add such a struct value as a `LuaRuntime` attribute, and store the pointer to that struct in `ud`.

scoder: We should then also include the initialisation phase in the memory counting, BTW, so that we cannot run into the case any more that anything gets deallocated that we didn't know about.

Le0Developer:
Test failure seems unrelated and can't reproduce

lupa/_lupa.pyx
```diff
@@ -460,6 +528,24 @@ cdef class LuaRuntime:
             lua.lua_settop(L, old_top)
             unlock_runtime(self)
 
+    def set_max_memory(self, size_t max_memory, count_base=False):
+        """Set maximum allowed memory for this LuaRuntime.
+
+        Setting max_memory to a value lower than currently in use, will set
+        max_memory to the current usage. Use strict=True to throw a
+        LuaMemoryError instead.
```
scoder, line 536: outdated

lupa/tests/test.py
```diff
@@ -3007,10 +3007,19 @@ def setUp(self):
             return self.skipTest("not supported in LuaJIT")
         return super(TestMaxMemory, self).setUp()
 
-    def test_property(self):
-        self.assertEqual(self.lua.max_memory, 10000)
+    def test_getters(self):
+        self.assertEqual(self.lua.get_memory_used(), 0)
+        self.assertGreater(self.lua.get_memory_used(True), 0)
```
scoder, line 3012: Please use keyword arguments instead of bare truth values to help readers understand what this does without looking up the function signature.

lupa/_lupa.pyx
```diff
@@ -265,12 +264,15 @@ cdef class LuaRuntime:
                   max_memory=None):
         cdef lua_State* L
 
-        self._memory_left = 0
-
         if max_memory is None:
             L = lua.luaL_newstate()
         else:
-            L = lua.lua_newstate(<lua.lua_Alloc>&_lua_alloc_restricted, &self._memory_left)
+            memory_status = <MemoryStatus*>calloc(1, sizeof(MemoryStatus))
```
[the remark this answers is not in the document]
Le0Developer: I allowed it to be NULL for easier checking if the lua state is using our custom allocator.

lupa/_lupa.pyx
```diff
@@ -276,16 +299,55 @@ cdef class LuaRuntime:
...
+
     def __dealloc__(self):
         if self._state is not NULL:
             lua.lua_close(self._state)
             self._state = NULL
+        if self._memory_status is not NULL:
+            free(self._memory_status)
+            self._memory_status = NULL
+
+    def get_max_memory(self, count_base=False):
```
scoder, line 318: To me, `count_base` reads like `base of count`, not `count the base`. Maybe `total` would be a better name? I think that's what it is, the total memory including the setup size of the runtime.

lupa/_lupa.pyx
```diff
@@ -276,17 +298,53 @@ cdef class LuaRuntime:
...
+        """
+        Maximum memory allowed to be used by this LuaRuntime.
+        0 indicates no limit meanwhile None indicates that the default lua
+        allocator is being used and ``set_max_memory()`` cannot be used.
+
+        If ``count_base`` is True, the base memory used by the lua runtime
+        will be included in the limit.
+        """
+        if self._memory_status.limit >= 0:
```
Le0Developer, line 327:
```suggestion
        if self._memory_status.limit < 0:
```

scoder: Right. Thanks for the fix.

Le0Developer:
I'm a Cython noob, so no idea how to fix this 🤷 
```cython
Error compiling Cython file:
------------------------------------------------------------
...
    cdef bytes _source_encoding
    cdef object _attribute_filter
    cdef object _attribute_getter
    cdef object _attribute_setter
    cdef bint _unpack_returned_tuples
    cdef MemoryStatus _memory_status
                     ^
------------------------------------------------------------

lupa/lua53.pyx:260:22: Variable type 'MemoryStatus' is incomplete
```

README.rst
```diff
@@ -913,6 +913,52 @@ setter function implementations for a ``LuaRuntime``:
...
+Restricting Lua Memory Usage
+----------------------------
+
+Lupa provides a simple mechanism to control the maximum memory
+usage of the Lua Runtime since version 2.0.
+By default Lupa does not interfere with Lua's memory allocation, to opt-in
+you must set the ``max_memory`` when creating the LuaRuntime.
+
+Lupa provides three helper functions for controlling and getting the
+memory usage of the LuaRuntime:
```
scoder, line 925:
```suggestion
The ``LuaRuntime`` provides three methods for controlling and reading the
memory usage:
```

README.rst
```diff
@@ -913,6 +913,52 @@ setter function implementations for a ``LuaRuntime``:
...
+
+2. ``get_max_memory(total=False)`` to get the current memory limit.
+   ``0`` means there is no memory limitation.
+
+3. ``set_max_memory(max_memory, total=False)`` to change the memory limit.
+   Values below or equal to 0 mean no limit.
+
+There is always some memory used by the LuaRuntime itself (around ~20KiB,
+depends on your lua version and other factors) which is excluded from all
+calculations unless specifying ``total=True``.
```
scoder, line 938:
```suggestion
There is always some memory used by the LuaRuntime itself (around ~20KiB,
depending on your lua version and other factors) which is excluded from all
calculations unless you specify ``total=True``.
```

README.rst
```diff
@@ -913,6 +913,52 @@ setter function implementations for a ``LuaRuntime``:
...
+depending on your lua version and other factors) which is excluded from all
+calculations unless you specify ``total=True``.
+
+.. code:: python
+
+        >>> lua = LuaRuntime(max_memory=0)  # 0 for unlimited, default is None
+        >>> lua.get_memory_used()  # memory used by your code
+        0
+        >>> lua.get_memory_used(total=True)  # includes memory used by the runtime itself
+        21358  # this will depend on your lua version and other factors
```
scoder, line 946:
```suggestion
        >>> total_lua_memory = lua.get_memory_used(total=True)  # includes memory used by the runtime itself
        >>> assert total_lua_memory > 0  # exact amount depends on your lua version and other factors
```

README.rst
```diff
@@ -913,6 +913,52 @@ setter function implementations for a ``LuaRuntime``:
...
+
+Lua code hitting the memory limit will receive memory errors:
+
+.. code:: python
+      
+        >>> lua.set_max_memory(100)
+        >>> lua.eval("string.rep('a', 1000)")
+        Traceback (most recent call last):
+         ...
+        lupa.LuaMemoryError: not enough memory
```
Le0Developer, line 957: This is throwing a `lupa.lua<version>.LuaMemoryError`, so the doctests fail here

scoder, line 957:
```suggestion
        >>> lua.eval("string.rep('a', 1000)")   # doctest: +ELLIPSIS
        Traceback (most recent call last):
         ...
        lupa...LuaMemoryError: not enough memory
```

Le0Developer:
All test succeeding locally now! 🎉

README.rst
```diff
@@ -913,6 +913,52 @@ setter function implementations for a ``LuaRuntime``:
...
+
+Lua code hitting the memory limit will receive memory errors:
+
+.. code:: python
+      
+        >>> lua.set_max_memory(100)
+        >>> lua.eval("string.rep('a', 1000)")   # doctest: +ELLIPSIS
+        Traceback (most recent call last):
+         ...
+        lupa...LuaMemoryError: not enough memory
```
scoder, line 957:
```suggestion
        ...LuaMemoryError: not enough memory
```

README.rst
```diff
@@ -913,6 +913,52 @@ setter function implementations for a ``LuaRuntime``:
...
+
+Lua code hitting the memory limit will receive memory errors:
+
+.. code:: python
+      
+        >>> lua.set_max_memory(100)
+        >>> lua.eval("string.rep('a', 1000)")   # doctest: +ELLIPSIS
+        Traceback (most recent call last):
+         ...
+        ...LuaMemoryError: not enough memory
```
scoder, line 957:
```suggestion
        >>> lua.eval("string.rep('a', 1000)")   # doctest: +IGNORE_EXCEPTION_DETAIL
        Traceback (most recent call last):
         ...
        lupa.LuaMemoryError: not enough memory
```

lupa/_lupa.pyx
```diff
@@ -460,6 +531,26 @@ cdef class LuaRuntime:
...
+        If ``total`` is True, the base memory used by the LuaRuntime itself
+        will be included in the memory limit.
+
+        If max_memory was set to None during creation, this will raise a
+        RuntimeError.
+        """
+        cdef size_t used
+        if self._memory_status.limit == <size_t> -1:
+            raise RuntimeError("max_memory must be set on LuaRuntime creation")
+        elif max_memory <= 0:
```
scoder, line 547: Since `max_memory` is a `size_t`, it cannot be `< 0`.
```suggestion
        elif max_memory == 0:
```

lupa/_lupa.pyx
```diff
@@ -460,6 +531,26 @@ cdef class LuaRuntime:
...
+            self._memory_status.limit = 0
+        elif total:
+            self._memory_status.limit = max_memory
+        else:
+            self._memory_status.limit = self._memory_status.base_usage + max_memory
```
scoder, line 552:
```suggestion
            self._memory_status.limit = self._memory_status.base_usage + max_memory
            # Prevent accidental (or deliberate) usage of our special value.
            if self._memory_status.limit == <size_t> -1:
                self._memory_status.limit -= 1
```

lupa/_lupa.pyx
```diff
@@ -276,17 +304,53 @@ cdef class LuaRuntime:
...
         self.init_python_lib(register_eval, register_builtins)
-        lua.lua_atpanic(L, <lua.lua_CFunction>1)
 
         self.set_overflow_handler(overflow_handler)
 
+        # lupa init done, set real limit
+        if max_memory is not None:
+            self._memory_status.base_usage = self._memory_status.used
+            if max_memory > 0:
+                self._memory_status.limit =  self._memory_status.base_usage + <size_t>max_memory
```
scoder, line 317:
```suggestion
                self._memory_status.limit =  self._memory_status.base_usage + <size_t>max_memory
                # Prevent accidental (or deliberate) usage of our special value.
                if self._memory_status.limit == <size_t> -1:
                    self._memory_status.limit -= 1
```

scoder:
Thanks!

PR #221, opened 2022-08-27
Make Lua 5.2 build

jmarianer:
For some reason, the makefile in the lua5-2 branch is missing a few
files. Rather than attempt to understand what happened there (my guess,
a bad import from some other VCS), this PR just adds those files in
Lupa's setup.py.


scoder:
Cool, thanks.

PR #224, opened 2022-09-22
Added option to pass already created LuaState

xthebat:
Hello, I stumble upon limitation of lupa library.

I some cases already created `lua_State` required to be used. But there is no possibility to pass pointer to `lua_State` in `LuaRuntime`. For example Wireshark has embedded Lua interpreter and this interpreter has bindings to Wireshark API. I wanted to write plugin (dissector) using Python because it is much more convenient than Lua for me (a lot libraries, familiar syntax, good IDE, etc). So it is not very hard to start Python interpreter using simple Lua C extensions module but without wrappers like lupa provided, I need to also write these wrappers by myself. So I wrote this simple patch to pass pointer of Lua interpreter into `LuaRuntime` and it works! I think may be it worth to be added to your library. 

Regards

### DIFF
--- a/lupa/_lupa.pyx
+++ b/lupa/_lupa.pyx
@@ -232,6 +232,7 @@ cdef class LuaRuntime:
       >>> lua_func(py_add1, 2)
       3
     """
+    cdef bint _lua_allocated
     cdef lua_State *_state
     cdef FastRLock _lock
     cdef dict _pyrefs_in_lua
@@ -243,11 +244,17 @@ cdef class LuaRuntime:
     cdef object _attribute_setter
     cdef bint _unpack_returned_tuples
 
-    def __cinit__(self, encoding='UTF-8', source_encoding=None,
+    def __cinit__(self, state=None, encoding='UTF-8', source_encoding=None,
                   attribute_filter=None, attribute_handlers=None,
                   bint register_eval=True, bint unpack_returned_tuples=False,
                   bint register_builtins=True, overflow_handler=None):
-        cdef lua_State* L = lua.luaL_newstate()
+
+        cdef lua_State *L = <lua_State*>state;
+        self._lua_allocated = False
+        if state is None:
+            self._lua_allocated = True
+            L = lua.luaL_newstate()
+
         if L is NULL:
             raise LuaError("Failed to initialise Lua runtime")
         self._state = L
@@ -276,14 +283,16 @@ cdef class LuaRuntime:
                 raise ValueError("attribute_filter and attribute_handlers are mutually exclusive")
             self._attribute_getter, self._attribute_setter = getter, setter
 
-        lua.luaL_openlibs(L)
+        if self._lua_allocated:
+            lua.luaL_openlibs(L)
+            
         self.init_python_lib(register_eval, register_builtins)
         lua.lua_atpanic(L, <lua.lua_CFunction>1)
 
         self.set_overflow_handler(overflow_handler)
 
     def __dealloc__(self):
-        if self._state is not NULL:
+        if self._state is not NULL and self._lua_allocated:
             lua.lua_close(self._state)
             self._state = NULL
 


Le0Developer:
Mind adding some tests?

scoder:
Thanks. I don't think it's that easy, though.

Lupa assumes that it owns the `lua_State`, and will deallocate it when the `LuaRuntime` is closed. It doesn't know that the `lua_State` is actually externally owned and that its lifetime may not be linked to the lifetime of the `LuaRuntime`. That's one issue.

Another thing is that `LuaRuntime.__init__()` does some setup of the Lua state and assumes that all of this is safe to do, because the Lua state was just created. That may be an unsafe assumption depending on where the `lua_State` originated from.

It's also unclear to me what you are passing into the constructor here. Is it the pointer address of the `lua_State` as a Python int? What if I pass `0` or `123` from Python?

I'm not generally opposed to such an interface, but then it should be provided as a C-level API and not be made publicly available to Python programs.

An example of your own usage could help to understand how you intend this to work.

xthebat:
Sorry for long response, I was bit busy on work.

@scoder 
> Lupa assumes that it owns the lua_State, and will deallocate it when the LuaRuntime is closed. It doesn't know that the lua_State is actually externally owned and that its lifetime may not be linked to the lifetime of the LuaRuntime. That's one issue.
> Another thing is that LuaRuntime.__init__() does some setup of the Lua state and assumes that all of this is safe to do, because the Lua state was just created. That may be an unsafe assumption depending on where the lua_State originated from.

I've added flag `self._lua_allocated` in 59a72a312c14c155d5af43e890932fec74b54de4 commit that indicate whether `lua_State` passed from outside or owned by `LuaRuntime`. In `__cinit__` and in `__dealloc__` I've added code to check for call `lua.luaL_openlibs` and `lua.lua_close` if `LuaRuntime` own pointer. It works for me, of course it doesn't mean it works everywhere :)

> It's also unclear to me what you are passing into the constructor here. Is it the pointer address of the lua_State as a Python int? What if I pass 0 or 123 from Python?

I've reworked it in 32fd682cf621fae7288d840af1be8bd249418546. Now `state` is Python int that should contain pointer to lua_State. Yeah, If we pass int that contains invalid pointer then program goes to segfault :( ruthless C world. In last commit no check was done to prevent this situation. I think this thought can help to avoid it for common Python user:

>  it should be provided as a C-level API and not be made publicly available to Python programs.

But all initialization LuaRuntime done in `__cinit__` and lua_State* should passed to it for correct initialization also for case external pointer. To hide such initialization variant in C-level I think LuaRuntime __cinit__ need to be somehow splitted.

> An example of your own usage could help to understand how you intend this to work.

When reworked my code I've added Cython proxy module in new commit 32fd682cf621fae7288d840af1be8bd249418546. You could find complete code in [embedded.pyx](https://github.com/scoder/lupa/pull/224/commits/32fd682cf621fae7288d840af1be8bd249418546#diff-746796b043e1f80ea425da47c85984673576e32e3c233806fdd856449b886d52) Also I've added code in `setup.py` to compile it. Compilation produce `embedded.cp39-win_amd64.pyd` (for win) and this library need to be placed in folder where target Lua interpreter looks for `require` C-modules and renamed to `libpylua.dll` (*.so if in Linux). In Lua:
```lua
require "libpylua"
python.import("some_package")
python.exec("some_var = some_package.somefunction()")
```
python.import added in 78f4b16d704d105422d511d51551afb3b66ea833

scoder:
> > it should be provided as a C-level API and not be made publicly available to Python programs.
> 
> But all initialization LuaRuntime done in `__cinit__` and lua_State* should passed to it for correct initialization also for case external pointer. To hide such initialization variant in C-level I think `LuaRuntime.__cinit__` need to be somehow splitted.

Idea: we can add an internal extension type (`@cython.internal`) that only wraps a `lua_State*` and can be passed as an optional argument, like this:
```cython
cdef class _ExternalLuaState:
    lua_State* lua_state

cdef api runtime_from_lua_state(lua_State* lua_state):
    cdef _ExternalLuaState arg = _ExternalLuaState.__new__(_ExternalLuaState)
    arg.lua_state = lua_state
    return LuaRuntime(lua_state=arg)
```

lupa/embedded.pyx
```diff
@@ -0,0 +1,88 @@
...
+    """
+    PyMODINIT_FUNC PyInit_embedded(void);
+
+    // use void* to make possible not to link proxy module with lua libraries
+    #define LUA_ENTRY_POINT(x) __declspec(dllexport) int luaopen_ ## x (void *L)
+
+    LUA_ENTRY_POINT(libpylua) {
+        PyImport_AppendInittab("embedded", PyInit_embedded);
+        Py_Initialize();
+        PyImport_ImportModule("embedded");
```
synodriver, line 76: I think we'd better add some checks here, e.g.
```
PyObject *mod = PyImport_ImportModule("embedded");
if (mod==NULL)
{
    PyErr_Print();
}
else
{
     Py_DECREF(mod);
}

PR #234, opened 2023-04-06
Add Lua 5.1.5 and 5.2.4 source code

xxyzz:
This pull request adds the Lua 5.1.5 and 5.2.4 source files. The Lua GitHub repo doesn't have tags for those two releases, so I downloaded the code from https://www.lua.org/ftp and added them to the `third-party` directory.

The [`test_decrease_memory`](https://github.com/scoder/lupa/blob/6813eee60fae1ac9a9f032aaaa27ea8d445d1751/lupa/tests/test.py#L3029-L3037) test failed on Lua 5.1 with error `not enough memory`, but I'm not sure how to fix it yet.


scoder:
Thanks. When you write "doesn't have tags", do you mean that it doesn't include the commit IDs either? Meaning, the code is really not there? Or are only the tags missing?

I'd rather not have these sources in the repo. If there's a way to get them via an external subrepo, I'd prefer that. Doesn't have to be github, just some sort-of official repo.

scoder:
At least for the 5.2 branch, there seem to be commits after 5.2.3:
https://github.com/lua/lua/commits/v5-2

The 5.1 branch has apparently been deleted already.

xxyzz:
The build finally passed this time... Only the `test_decrease_memory` test failed for Lua 5.1.

setup.py
```diff
@@ -263,6 +263,9 @@ def use_bundled_lua(path, macros):
 
     # Parse .o files from 'makefile'
     makefile = os.path.join(path, "makefile")
+    if 'lua51' in path:
+        # Lua 5.1.5 source code is downloaded from lua.org/ftp
+        makefile = os.path.join(path, "src", "Makefile")
```
scoder, line 268: Can't we have a list of places where we look, and take the first file that we find? I don't think "lua51" is special enough to break the rules here. And it might even fail to do the right thing if users name their working/venv directory "lua51_52_53", for example (because then "lua51" is in *every* path).

xxyzz: How about checking the `libname` variable:

```python
if libname == 'lua51':
```

setup.py
```diff
@@ -263,6 +263,9 @@ def use_bundled_lua(path, macros):
 
     # Parse .o files from 'makefile'
     makefile = os.path.join(path, "makefile")
+    if libname == 'lua51':
+        # Lua 5.1.5 source code is downloaded from lua.org/ftp
+        makefile = os.path.join(path, "src", "Makefile")
```
scoder, line 268:
```suggestion
    # Find Makefile in subrepos and downloaded sources.
    for makefile_path in [os.path.join("src", "makefile"), os.path.join("src", "Makefile"), "makefile", "Makefile"]:
        makefile = os.path.join(path, makefile_path)
        if os.path.exists(makefile):
            break
    else:
        raise RuntimeError("Makefile not found in " + path)
```

xxyzz: That would fail on Windows. Because Windows path is case insensitive, and lua_5.1.5.tar.gz also has a `Makefile` in the root directory but that file doesn't have any `.o` target.

scoder:
Thanks!

rsiemens:
> Personally, I think that Lua 5.1 is long out of maintenance for good reason, and shouldn't be used any more, but if we can still support it without much effort, then let's just add the sources to Lupa. It just makes the distro and repo a little larger. Someone out there can probably still make use of it.

Just another supporting data point for Lua 5.1 inclusion - [Redis still bundles Lua 5.1](https://redis.io/docs/interact/programmability/lua-api/) so being able to easily unit test any lua scripts that will be sent to redis via python is nice.

PR #237, opened 2023-04-18
Fix bug when installing with `--no-bundle`

riconnon:
Wrap the result of finding the system lua version in an array to be compatible with the array of bundled configurations introduced in 2.0


scoder:
Thanks. There was also a second case which I fixed in https://github.com/scoder/lupa/commit/19279acda1ad7e7a536adafe399b183701287bc1

PR #238, opened 2023-04-27
Add pyproject.toml build file

xxyzz:
This fixes the pip install error when adding Lupa source code as dependency in other packages' pyproject.toml file because Lupa's `setup.py` doesn't install Cython automatically.

Example `pyproject.toml`:

```
[build-system]
requires = ["setuptools"]
build-backend = "setuptools.build_meta"

[project]
name = "mypackage"
version = "0.0.1"
dependencies = [
    "lupa @ git+https://github.com/scoder/lupa.git"
]
```

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -491,7 +491,7 @@ setup(
     ],
 
     packages=['lupa'],
-    build_requires=['Cython>=0.29.28'],
+    setup_requires=['Cython>=0.29.28'],
     ext_modules=ext_modules,
     libraries=ext_libraries,
     **extra_setup_args


xxyzz:
Hi @scoder, could you please review this minor change if you have the time?

xxyzz:
But setuptools doesn't recognize the `build_requires` keyword. It has [`setup_requires` and `install_requires`](https://setuptools.pypa.io/en/latest/references/keywords.html), but that would install Cython as normal dependency instead of build dependency package and Cython is not installed when `prepare_extensions()` is running according to the build log.

scoder:
> But setuptools doesn't recognize the `build_requires` keyword.

Understood. so that can probably be removed. Still, the GHA and appveyor scripts both use `requirements.txt`, so that is the single point of truth. Is there a way to make `pyproject.toml` use `requirements.txt`?

xxyzz:
>  Is there a way to make `pyproject.toml` use `requirements.txt`?

Probably not, `project.dependencies` can use [dynamic metadata](https://setuptools.pypa.io/en/latest/userguide/pyproject_config.html#dynamic-metadata) to set a file but not for `build-system.requires`.

xxyzz:
I removed the `pyproject.toml` file and replaced the invalid `build_requires` keyword with `setup_requires`. Now Cython can be installed with pip install's `--use-pep517` option.

scoder:
Thanks

Issue #235, opened 2023-04-16
setup.py does not build with luajit on windows(x64)


hidwood:
`This` issue occurs when building with the standard "setup.py build" command, as it completely skips checking and building LuaJit. 
It succeeds in building Lua51, Lua52, Lua53, and Lua54, and then skips both LuaJit20 and LuaJit21, despite being  able to build them directly. 
After searching the setup.py, I found a comment that states:
#Couldn't get the Windows build to work. See
            #https://luajit.org/install.html#windows
            I am unsure if this means that the setup file cannot build on windows, or if it means that it is expecting one to build it themselves. Just in case that was true, I built with "msvcbuild.bat", and reran, but nothing had changed. After more searching, it does not appear that this should be the case, as there are references to msvcbuild in the setup file. I am now at a loss and would greatly appreciate some insight into what is going on. Fixing this issue would allow for LuaJit to be distributed in the auto-generated wheels for windows, much like it already is in Linux.

MuffinMario:
It seems that luajit is currently not working with windows as it is written. What worked for me was the following:

1. Go to setup.py and comment out the previously mentioned line `or (platform.startswith('win') and 'luajit' in os.path.basename(lua_bundle_path.rstrip(os.sep)))`
2. Modify the line below which is limiting luajit to only x64 arch so that it also supports windows (which [for some reason](https://stackoverflow.com/questions/45124888/possible-values-for-platform-machine) says "AMD64" instead of "x64_86") `or (get_machine() != "AMD64" and  get_machine() != "x86_64" and 'luajit' in os.path.basename(lua_bundle_path.rstrip(os.sep)))`
3. Open your Visual Studio 20XX x64 native command prompt (or alternatively run vsvars64.bat which for 2022 is lying in C:\Program Files\Microsoft Visual Studio\2022\Community\VC\Auxiliary\Build in **cmd** as powershell will not work)
4. `cd` to your directory and run `python setup.py build_ext -i`

This will include the luajit20 and luajit21 third-party folders to also build, leaving you with lua52,lua53,lua54,luajit20 and luajit21 packages.

scoder:
If someone gets this to work, I'd be happy to receive a PR.

hidwood:
Thank you, this   appears to work. I suppose my question is then why Windows on x64 will not build unless the machine type is specifically not x64 or AMD64? One would think that it would, but I do not know enough about the wizardry that is going on to make a guess as to why it does not work by default. 
[...]

PR #251, opened 2023-12-01
allows for proper building of LuaJit on Windows x64

hidwood:
This pull request makes slight changes to the currently existing setup.py to bypass the check for the name of the platform to start with 'win'. This allows LuaJIT to be built successfully on my machines, two running Windows 10 and another running Ubuntu 20.04. I have no way to test this on a Mac, however I do not believe it should impact it in any way. The only potential issue I could see with this fix is one mentioned in #235 where the CI may fail due to a hardcoded Unix path. However, as I am currently not aware of how the directory structure may change under a windows build host, I am unable to make this alteration myself. If anyone knows if there is a change in the directory structure, I would be happy to implement that here as well as to not clutter the repository with merges.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -228,6 +228,11 @@ def use_bundled_luajit(path, macros):
     build_env = dict(os.environ)
     src_dir = os.path.join(path, "src")
     if platform.startswith('win'):
+        try:
+            subprocess.run(["C:\\Program Files(x86)\\Microsoft Visual Studio\\2019\\Community\\VC\\Auxiliary\\Build\\vcvarsall.bat", "x86_x64",])
+        except FileNotFoundError:
+            print("vcvarsall not found in the location spesified by appveyor.yml. If you are not building on CI, this can safely be ignored.")
+
         build_script = [os.path.join(src_dir, "msvcbuild.bat"), "static"]
         lib_file = "lua51.lib"
     else:
--- a/setup.py
+++ b/setup.py
@@ -365,10 +365,6 @@ if not configs and not option_no_bundle:
             # http://t-p-j.blogspot.com/2010/11/lupa-on-os-x-with-macports-python-26.html
             # LuaJIT 2.1-alpha3 fails at runtime.
             or (platform == 'darwin' and 'luajit' in os.path.basename(lua_bundle_path.rstrip(os.sep)))
-            # Couldn't get the Windows build to work. See
-            # https://luajit.org/install.html#windows
-            or (platform.startswith('win') and 'luajit' in os.path.basename(lua_bundle_path.rstrip(os.sep)))
-            # Let's restrict LuaJIT to x86_64 for now.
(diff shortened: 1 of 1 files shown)


scoder:
The Windows build fails because we don't call `msvcbuild.bat` correctly from a VS prompt. I don't know how to do that. Any idea?

scoder:
This explains it:
https://learn.microsoft.com/en-us/visualstudio/ide/reference/command-prompt-powershell?view=vs-2022#start-from-file-browser
However, it seems to suggest figuring out the absolute path to a VS batch script. :-/

scoder:
It seems that `vcvarsall.bat` is what we want here:
https://learn.microsoft.com/en-us/cpp/build/building-on-the-command-line?view=msvc-170

hidwood:
My apologies, I had been running from vcvars64 when I did my tests as that is usually something that one does when building cython projects on windows in my experience. I am not sure how one would find the location of vcvars64 or vcvarsall to build against however as it seems like it could be version specific. Additionally, I am unsure about the Visual Studio installation location on the CI.
By default, vcvarsall is found in "C:\Program Files\Microsoft Visual Studio<vs_version><variant>\VC\Build," where `<vs_version>` is something like 2022 or 2019, and `<variant>` can be community, team, or enterprise. Additionally, it does not appear that an environment variable is created or added to path that would give the possible location of Visual studio. Without knowing this, I am not entirely sure how one could find it without hardcoding the path and hoping for the best, which is obviously not an optimal solution.

hidwood:
I have added a check that searches for where I believe the file should be in the currently selected appveyor image. from what little I could find by looking through the install script. I found the following for reference:
""${env:ProgramFiles(x86)}\Microsoft Visual Studio\2019\Community""
When converted to a standard windows path that translates to "C:\Program Files (x86)\Microsoft Visual Studio\2019\Community"
Hopefully this fixes the build issue, although I am not sure if ${env:ProgramFiles(x86)} mapps to somewhere else.

setup.py
```diff
@@ -228,6 +228,11 @@ def use_bundled_luajit(path, macros):
     build_env = dict(os.environ)
     src_dir = os.path.join(path, "src")
     if platform.startswith('win'):
+        try:
+            subprocess.run(["C:\\Program Files(x86)\\Microsoft Visual Studio\\2019\\Community\\VC\\Auxiliary\\Build\\vcvarsall.bat", "x86_x64",])
```
scoder, line 232: I don't think it helps to do this. IIUC, the batch file sets up environment variables, so it needs to run in the same command shell that then executes the build.

Also, I don't think it's a good idea to use a hard coded path. There must be a way to figure it out programmatically. It might also be that the batch file is simply on the search PATH already and can be started as is.

Note that appveyor isn't really used any more since we use GHA also for the Windows builds. I should disable it.

Issue #248, opened 2023-10-28
Expose Lua load arguments (mode and chunkname)


astoff:
It would be nice to expose the mode and chunkname arguments of Lua's load function in `LuaRunner.{eval,execute,compile}`.

Setting mode to `"t"` is important for sanboxing purposes. Specifying the chunk name is less important by I guess it's nice to have.

gentooise:
This is a patch that I wrote to expose the chunk name (it works for my needs). I don't have time right now to open a PR, but I'll leave the changes here in case anyone needs them:

```patch
diff --git a/lupa/_lupa.pyx b/lupa/_lupa.pyx
index 9310e35..34a4d3b 100644
--- a/lupa/_lupa.pyx
+++ b/lupa/_lupa.pyx
@@ -393,36 +393,45 @@ cdef class LuaRuntime:
             raise
         return 0

-    def eval(self, lua_code, *args):
+    def eval(self, lua_code, name = '<python>', *args):
         """Evaluate a Lua expression passed in a string.
         """
         assert self._state is not NULL
         if isinstance(lua_code, unicode):
             lua_code = (<unicode>lua_code).encode(self._source_encoding)
-        return run_lua(self, b'return ' + lua_code, args)
+        if not isinstance(name, (bytes, unicode)):
+            raise TypeError("Lua program name must be a string")
+        name = (<unicode>name).encode('UTF-8')
+        return run_lua(self, b'return ' + lua_code, name, args)

-    def execute(self, lua_code, *args):
+    def execute(self, lua_code, name = '<python>', *args):
         """Execute a Lua program passed in a string.
         """
         assert self._state is not NULL
         if isinstance(lua_code, unicode):
             lua_code = (<unicode>lua_code).encode(self._source_encoding)
-        return run_lua(self, lua_code, args)
+        if not isinstance(name, (bytes, unicode)):
+            raise TypeError("Lua program name must be a string")
+        name = (<unicode>name).encode('UTF-8')
+        return run_lua(self, lua_code, name, args)

-    def compile(self, lua_code):
+    def compile(self, lua_code, name = '<python>'):
         """Compile a Lua program into a callable Lua function.
         """
         assert self._state is not NULL
         cdef const char *err
         if isinstance(lua_code, unicode):
             lua_code = (<unicode>lua_code).encode(self._source_encoding)
+        if not isinstance(name, (bytes, unicode)):
+            raise TypeError("Lua program name must be a string")
[...]

PR #252, opened 2023-12-10
Add 'name' and 'mode' arguments to LuaRuntime.{eval,execute,compile}

scoder:
These allow finer control of input and debug output.

Closes https://github.com/scoder/lupa/issues/248

### DIFF
--- a/lupa/_lupa.pyx
+++ b/lupa/_lupa.pyx
@@ -393,36 +393,74 @@ cdef class LuaRuntime:
             raise
         return 0
 
-    def eval(self, lua_code, *args):
+    @cython.final
+    cdef bytes _source_encode(self, string):
+        if isinstance(string, unicode):
+            return (<unicode>string).encode(self._source_encoding)
+        elif isinstance(string, bytes):
+            return <bytes> string
+        elif isinstance(string, bytearray):
+            return bytes(string)
+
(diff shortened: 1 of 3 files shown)


lupa/_lupa.pyx
```diff
@@ -1719,14 +1757,21 @@ cdef build_lua_error_message(LuaRuntime runtime, lua_State* L, int stack_index=-
 
 # calling into Lua
 
-cdef run_lua(LuaRuntime runtime, bytes lua_code, tuple args):
+cdef run_lua(LuaRuntime runtime, bytes lua_code, bytes name, bytes mode, tuple args):
     """Run Lua code with arguments"""
     cdef lua_State* L = runtime._state
+    cdef const char* c_name = b'<python>'
```
astoff, line 1763: This line I would change to

```suggestion
    cdef const char* c_name = b'=<python>'
```
or

```suggestion
    cdef const char* c_name = NULL
```
for the reason explained [here](https://github.com/scoder/lupa/issues/248#issuecomment-1822909833).

scoder: I've seen your comment but it would introduce an incompatible change. It would break doctests, for example, or code that tries to parse Lua errors.

Issue #250, opened 2023-11-03
Deadlock involving __dealloc__


astoff:
I'm facing a deadlock which I think cannot be avoided from the Python API alone. The situation is as follows:

1. A `LuaRuntime` is running on thread B.
2. It calls a Python function. The argument is a Lua object.
3. The Lua object gets diligently converted to a pure-Python data structure.
4. The Python data structure is passed to thread A.
5. Thread B blocks until thread A returns something.

I've traced the execution and the deadlock happens on this line of `_LuaObject.__dealloc__`:

https://github.com/scoder/lupa/blob/67ba19406d6587139fc2779f549e9543da68ff11/lupa/_lupa.pyx#L823

My assumption is that the "diligent conversion" step 3 generates some garbage that gets collected a bit later, when thread A is running. As evidence to this theory:

* I haven't observed a deadlock yet if I call `gc.collect(0)` between steps 3 and 4.
* I observe no deadlock if I remove the `__dealloc__` method.

(As a data point for you, in case it is hard to find a fully general solution to this issue, my Lua runs are realtively short and I don't mind imperfect garbage collection as in the second point above.)

scoder:
Hmm, interesting case. That makes locking the runtime appear much less appealing inside of `__dealloc__`. It's difficult to avoid, though.

The lock that Lupa uses is re-entrant, so this rarely poses problems. It probably requires a setup as in your case to trigger it: passing Python objects over to other Lua threads inside of a Lua call.

One possible solution that comes to my mind is to request the lock in a non-blocking way in cases where we cannot afford waiting (in `__dealloc__` specifically), and if we fail to acquire it, add the object to a list in the `LuaRuntime` and deallocate that list the next time right before we release the lock (i.e. when we definitely own it any way). That would probably keep some objects alive longer than necessary, and there might also be situations where this has other unexpected effects (we're dealing with threads here, so anything can happen, really), but it should still work as before in most cases and only behave differently in cases where it's currently dead-lock prone already. And probably provides an improvement in that case.

astoff:
I agree with what you wrote, and I can contribute a test case which currently fails for me:

```python
def test_lupa_gc_deadlock():

    def assert_no_deadlock(thread):
        thread.start()
        thread.join(1)
        assert not thread.is_alive(), "thread didn't finish"

    def trigger_gc(ref):
        del ref[0]

    lua = LuaRuntime()
    ref = [lua.eval("{}")]
    lua.execute(
[...]

PR #255, opened 2024-01-28
Try to solve deadlock when deallocating Python objects in threads

scoder:
Closes https://github.com/scoder/lupa/issues/250


astoff:
I suppose the issue here is in the test itself, as this change makes it pass on my local machine:
```diff
modified   lupa/tests/test.py
@@ -103,14 +103,13 @@ class TestLuaRuntimeRefcounting(LupaTestCase):
         self._run_gc_test(make_refcycle, off_by_one=True)
 
     def test_lupa_gc_deadlock(self):
-        lua = self.lupa.LuaRuntime()
-
         def assert_no_deadlock(thread):
             thread.start()
             thread.join(1)
             assert not thread.is_alive(), "thread didn't finish - deadlock?"
 
         def delete_table_reference_in_thread():
+            lua = self.lupa.LuaRuntime()
             ref = [lua.eval("{}")]
 
             def trigger_gc(ref):
```

astoff:
@scoder Did you get the chance to try out my previous comment from last month?

scoder:
I had looked at it when you wrote it and wasn't sure. The difference is that, in your proposal, the Lua runtime gets cleaned up after each run, thus clearing all objects and freeing all memory. It's good to know that that doesn't leave anything behind, but my implementation wanted to make sure that the objects get cleaned up as early as possible (and safe), preferably before returning from Lua to Python, and not just kept around unnecessarily. Your change doesn't test that any more.

I still think there's something wrong (or hopefully only missing) in my implementation.

scoder:
I took another look and it turns out that most of the excess references in the test are cleaned up by triggering a run of the Lua garbage collector. Most, but not all. There's around 20 remaining references that are created by the test and not cleaned up afterwards:
```
[(<cell at 0x7fbd0f040bb0: dict object at 0x7fbd0f1d2500>,),
 <function TestLuaRuntimeRefcounting._run_gc_test.<locals>.<listcomp> at 0x7fbd0f0b8790>,
 <lupa.luajit20._PyReference object at 0x7fbd0f041030>,
 <weakref at 0x7fbd0f0df7e0; to 'Thread' at 0x7fbd0f0408e0>,
 <lupa.luajit20._PyReference object at 0x7fbd0f040a00>,
 <Thread(Thread-707 (trigger_gc), stopped 140449962329664)>,
 {'_daemonic': False,
  '_ident': 140449962329664,
  '_initialized': True,
  '_invoke_excepthook': <function _make_invoke_excepthook.<locals>.invoke_excepthook at 0x7fbd0ef500d0>,
  '_is_stopped': True,
  '_name': 'Thread-707 (trigger_gc)',
  '_native_id': 134030,
  '_started': <threading.Event object at 0x7fbd0f040f40>,
  '_stderr': <_io.TextIOWrapper name='<stderr>' mode='w' encoding='utf-8'>,
  '_tstate_lock': None},
 <threading.Event object at 0x7fbd0f040f40>,
 <function _make_invoke_excepthook.<locals>.invoke_excepthook at 0x7fbd0ef500d0>,
 {'_cond': <Condition(<unlocked _thread.lock object at 0x7fbd0f1e3e40>, 0)>, '_flag': True},
 (<cell at 0x7fbd0f0407f0: builtin_function_or_method object at 0x7fbd0f6b9530>,
  <cell at 0x7fbd0f040490: module object at 0x7fbd0f6a6390>,
  <cell at 0x7fbd0f041150: builtin_function_or_method object at 0x7fbd0f6d6110>,
  <cell at 0x7fbd0f0410c0: function object at 0x7fbd0f5afc70>,
  <cell at 0x7fbd0f040b20: builtin_function_or_method object at 0x7fbd0f6a66b0>),
 <Condition(<unlocked _thread.lock object at 0x7fbd0f1e3e40>, 0)>,
 <cell at 0x7fbd0f040b20: builtin_function_or_method object at 0x7fbd0f6a66b0>,
 <cell at 0x7fbd0f0410c0: function object at 0x7fbd0f5afc70>,
 <cell at 0x7fbd0f041150: builtin_function_or_method object at 0x7fbd0f6d6110>,
 <cell at 0x7fbd0f040490: module object at 0x7fbd0f6a6390>,
 <cell at 0x7fbd0f0407f0: builtin_function_or_method object at 0x7fbd0f6b9530>,
 {'_lock': <unlocked _thread.lock object at 0x7fbd0f1e3e40>,
  '_waiters': deque([]),
  'acquire': <built-in method acquire of _thread.lock object at 0x7fbd0f1e3e40>,
  'release': <built-in method release of _thread.lock object at 0x7fbd0f1e3e40>},
 <unlocked _thread.lock object at 0x7fbd0f1e3e40>,
 <built-in method acquire of _thread.lock object at 0x7fbd0f1e3e40>,
 <built-in method release of _thread.lock object at 0x7fbd0f1e3e40>,
 deque([])]
```
The test passes the Thread object through Lua and it's not obvious if some of these are related to that.

astoff:
> my implementation wanted to make sure that the objects get cleaned up as early as possible (and safe), 

I don't know much about GC, but isn't this due to Lua not using a reference count approach? What I imagine is, with reference counts you can deallocate objects as soon as the count reaches 0, but with other GC approaches every object survives at least until the next GC cycle runs.

>  ... preferably before returning from Lua to Python

In other words, in Lua you wouldn't expect that GC always runs before returning to Python.

Anyway, this is just a wild hunch. I went through the old version of this PR and couldn't spot anything missing in it.

scoder:
I think it's too difficult to test correctly for reference leaks here, so the test sadly has to go without that. I simplified it.

PR #279, opened 2025-02-15
remove IS_PYTHON2 from tests too

a-detiste:
more cleanup :-)

### DIFF
--- a/lupa/tests/test.py
+++ b/lupa/tests/test.py
@@ -161,13 +161,12 @@ class TestLuaRuntime(SetupLuaRuntimeMixin, LupaTestCase):
     def test_eval_error_message_decoding(self):
         try:
             self.lua.eval('require "UNKNOWNöMODULEäNAME"')
-        except self.lupa.LuaError:
-            error = str(sys.exc_info()[1])
+        except self.lupa.LuaError as exc:
+            error = str(exc)
         else:
             self.fail('expected error not raised')
         expected_message = 'module \'UNKNOWNöMODULEäNAME\' not found'
-        self.assertTrue(expected_message in error,
-                        '"%s" not found in "%s"' % (expected_message, error))
+        self.assertIn(expected_message, error)
 
     def test_execute(self):
         self.assertEqual(2, self.lua.execute('return 1+1'))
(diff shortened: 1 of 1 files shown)


lupa/tests/test.py
```diff
@@ -170,12 +162,10 @@ def test_eval_error_message_decoding(self):
         try:
             self.lua.eval('require "UNKNOWNöMODULEäNAME"')
         except self.lupa.LuaError:
-            error = ('%s'.decode('ASCII') if IS_PYTHON2 else '%s') % sys.exc_info()[1]
+            error = '%s'.decode('ASCII') % sys.exc_info()[1]
```
scoder, line 165:
```suggestion
            error = str(sys.exc_info()[1])
```

lupa/tests/test.py
```diff
@@ -2091,10 +2079,7 @@ def mandelbrot(i, lua_func):
         # plausability checks - make sure it's not all white or all black
         self.assertEqual('\0'.encode('ASCII')*(image_size//8//2),
                          result_bytes[:image_size//8//2])
-        if IS_PYTHON2:
-            self.assertTrue('\xFF' in result_bytes)
-        else:
-            self.assertTrue('\xFF'.encode('ISO-8859-1') in result_bytes)
+        self.assertTrue('\xFF'.encode('ISO-8859-1') in result_bytes)
```
scoder, line 2082:
```suggestion
        self.assertTrue(b'\xFF' in result_bytes)
```

lupa/tests/test.py
```diff
@@ -162,7 +162,7 @@ def test_eval_error_message_decoding(self):
         try:
             self.lua.eval('require "UNKNOWNöMODULEäNAME"')
         except self.lupa.LuaError:
-            error = '%s'.decode('ASCII') % sys.exc_info()[1]
+            error = str(sys.exc_info()[1])
```
a-detiste, line 165: exc_info is already always a `str`

scoder: No, it's an exception instance. But looking at it again, I noticed that we can avoid calling `sys.exc_info` all together just to get at the current exception that we're handling.

a-detiste: I'm happy this goes well. Some other projects want to keep their cargo-culted `compat.py` forever.
There's always opportunity to learn new tricks.

scoder:
Thanks

PR #280, opened 2025-02-15
remove more traces of Python2

a-detiste:


scoder (commented):
Looks good to me.

The `_thread` import is probably something to avoid but that's outside of the scope of this PR.

scoder:
Thanks

Issue #294, opened 2026-08-13
Python userdata identity can get mixed up in Lua 5.5


stelonix:
I think I've run into a Python userdata/reference issue with the new lua 5.5 backend.

This reproduces for me with lupa 2.8:

```python
import lupa.lua54 as lua54
import lupa.lua55 as lua55


class Entity:
    def __init__(self, name):
        self.name = name

    def __repr__(self):
        return f"Entity({self.name})"


def test(lua):
    runtime = lua.LuaRuntime()
    echo = runtime.eval("function(entity, event) return entity end")

    entities = [Entity("A"), Entity("B"), Entity("C")]

    for i in range(10000):
        expected = entities[i % 3]
        actual = echo(expected, runtime.table(name="update"))

        if actual is not expected:
            print("failed at", i)
            print("expected:", expected, id(expected))
            print("actual:  ", actual, id(actual))
            return

    print("passed")


print("lua54:")
test(lua54)

print("lua55:")
test(lua55)
```

For me:

```text
lua54:
passed

lua55:
failed at 51
expected: Entity(A)
actual:   Entity(C)
```

The exact failing call can move around in a larger repro, but lua55 consistently ends up substituting another live Python object for the one that was passed in. lua54 survives 10,000 calls.

I originally found this in a Lua -> Python -> Lua callback path where an entity argument eventually became another Python object associated with that entity (a dict). I reduced it down to the above, which doesn't need the callback/re-entry at all.

Possibly related to #293, although that one reproduces on 5.2/5.3 and not 5.4+, while this one seems specific to 5.5 for me.

Environment:

```text
Lupa 2.8
Python 3.10.12
Ubuntu 22.04 x86_64
Linux 6.8.0-136-generic #136~22.04.1-Ubuntu SMP PREEMPT_DYNAMIC Fri Jul 3 16:29:11 UTC 2026
```

I also checked the installed extensions: `lupa.lua54` and `lupa.lua55` are separate binaries and neither dynamically links a system liblua, so this doesn't look like a 5.4/5.5 shared-library ABI mixup.

For now I've fixed it by using the 5.4 backend.

rbroderi:
I was able to reproduce as well.
```python
"""Minimal standalone repro for Lupa Python-object identity corruption."""

import importlib

from typing import Any

from typing import cast
 

_ITERATIONS = 20_000
 

def _run(name: str) -> None:
    try:
        lua_module = importlib.import_module(name)
    except ImportError:
[...]

PR #295, opened 2026-09-11
Fix for issues/294

rbroderi:
A possible fix for https://github.com/scoder/lupa/issues/294 I won't claim to be an amazing c programmer but this fix seems to be where the problem is. I would suggest reviewing this as it was AI assisted. It would be at least a good start as to a fix for https://github.com/scoder/lupa/issues/294.


scoder:
Thanks. So, from what I understand, the issue is that Lua's garbage collection (in Lua 5.5) can take its time to free unused userdata entries after Lupa marked them as released, and Lupa might end up reusing the slot before it's actually freed.

You resolve that with a freelist that keeps all userdata alive and never frees it, and instead does its own reuse management. Right?

That doesn't seem a bad idea, and it might even have a positive effect on allocation speed. But it does have the downside of keeping all userdata alive, which can become a visible memory leak when memory usage (or, rather, the live object count) changes largely over time.

I think I found a simpler fix in https://github.com/scoder/lupa/pull/296

Please test it on your side. I tried to take care that it doesn't introduce memory leaks but I'm not entirely sure that I succeeded.

rbroderi:
Thanks for the quick look at this. I will check it out. And yes your fix does seem simpler.

